PR #2, opened 2018-11-29
Fix symlink issues and Windows support

svanharmelen:
These changes:

- ignore the `.git` directory
- ignore the `.terraform` directory, except for the `.terraform/modules` subdirectory
- make all symlinks within the directory structure that is packed relative
- dereference all symlinks to files or directories outside the directory structure that is packed
- preserve relative symlinks within dereferenced folders


slug.go
```diff
@@ -18,31 +19,51 @@ type Meta struct {
 	Size int64
 }
 
-// Pack creates a slug from a directory src, and writes the new
-// slug to w. Returns metadata about the slug and any error.
-func Pack(src string, w io.Writer) (*Meta, error) {
-	// Gzip compress all the output data
+// Pack creates a slug from a src directory, and writes the new slug
+// to w. If dereference is set to true, symlinks with a target ouside
+// of the src directory will be dereferenced. Returns metadata about
```
ryanuber: If `dereference` is false, then we omit the symlink completely if I am reading this right. That's fine! But maybe we should call that out here to be clear.

svanharmelen: That is correct, I'll make that more explicit in the comment 👍

PR #3, opened 2019-04-03
Update the filter logic to be more precise

svanharmelen:


slug.go
```diff
@@ -75,11 +75,20 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
 			return nil
 		}
 
-		// Skip the .terraform directory, except for the modules subdirectory.
-		if strings.Contains(subpath, ".terraform") && info.Name() != ".terraform" {
-			if !strings.Contains(subpath, filepath.Clean(".terraform/modules")) {
-				return filepath.SkipDir
-			}
+		// Ignore the .terraform directory.
+		if info.Name() == ".terraform" {
```
ryanuber: How about adding an IsDir() here?

slug.go
```diff
@@ -75,11 +75,20 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
...
+			return nil
+		}
+
+		// Ignore any files in the .terraform directory.
+		if !info.IsDir() && filepath.Dir(subpath) == ".terraform" {
+			return nil
+		}
+
+		// Skip .terraform subdirectories, except for the modules subdirectory.
+		if strings.Contains(subpath, ".terraform/") &&
```
ryanuber: We should probably use `strings.HasPrefix()` for this right? As-is, this would catch stuff like `foo.terraform/bar`.

slug.go
```diff
@@ -75,11 +75,20 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
...
+			!strings.Contains(subpath, filepath.Clean(".terraform/modules")) {
```
ryanuber: The `filepath.Clean` is probably unnecessary since it's hardcoded, no?

svanharmelen: This is needed to make sure that the path separator is correct for the OS (Windows/Linux) that is being used.

slug.go
svanharmelen: Yep, good point 👍

slug.go
svanharmelen: I think it still used `Contains` based on some earlier tests around this logic. But indeed we can now use `HasPrefix()` here to make it more robust 👍 Added a test for it as well.

PR #5, opened 2019-08-30
Handle multiple root modules correctly

0gajun:
In Terraform Enterprise, repositories can contain multiple terraform
root modules. Like following.

```
/ - project1/ - main.tf
  - project2/ - main.tf
```

With this repository structure, the current implementation cannot filter
out `project1/.terraform/plugins` and `project2/.terraform/plugins`,
because `.terraform` directories aren't located on a repository root.
This change filters out such directories and makes the remote plan
faster by reducing data transfer.

In our case, `.terraform/plugins` directories contain aws provider binaries.
The binaries are so large about 200MB per binary.
Our single repository has about 8 root modules, so sometimes we transfer more than 1.5GB to Terraform Enterprise in remote plan 😱

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -81,13 +81,14 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
 		}
 
 		// Ignore any files in the .terraform directory.
-		if !info.IsDir() && filepath.Dir(subpath) == ".terraform" {
+		if !info.IsDir() && filepath.Base(filepath.Dir(subpath)) == ".terraform" {
 			return nil
 		}
 
 		// Skip .terraform subdirectories, except for the modules subdirectory.
-		if strings.HasPrefix(subpath, ".terraform"+string(filepath.Separator)) &&
-			!strings.HasPrefix(subpath, filepath.Clean(".terraform/modules")) {
+		if info.IsDir() &&
+			filepath.Base(filepath.Dir(subpath)) == ".terraform" &&
+			info.Name() != "modules" {
 			return filepath.SkipDir
 		}
 
--- a/slug_test.go
+++ b/slug_test.go
@@ -73,8 +73,10 @@ func TestPack(t *testing.T) {
 	// Make sure the .terraform directory is ignored,
 	// except for the .terraform/modules subdirectory.
 	for _, file := range fileList {
-		if strings.HasPrefix(file, ".terraform"+string(filepath.Separator)) &&
-			!strings.HasPrefix(file, filepath.Clean(".terraform/modules")) {
+		dirPath := filepath.Dir(file)
+		parentDirPath := filepath.Dir(dirPath)
+		if filepath.Base(parentDirPath) == ".terraform" &&
+			filepath.Base(dirPath) != "modules" {
 			t.Fatalf("unexpected .terraform content: %s", file)
 		}
 	}
--- a/testdata/archive-dir/sub/.terraform/modules/README
+++ b/testdata/archive-dir/sub/.terraform/modules/README
@@ -1,0 +1,2 @@
+Keep this file and directory here to test if its properly ignored
+
--- a/testdata/archive-dir/sub/.terraform/plugins/README
+++ b/testdata/archive-dir/sub/.terraform/plugins/README
@@ -1,0 +1,2 @@
+Keep this file and directory here to test if its properly ignored
+


svanharmelen:
I see what you are trying to fix and I think it's something worth exploring, but unfortunately the currently suggested solution doesn't work with a standard Terraform setup and I even think it will break things.

Let's assume the same repo you described with two projects:

```
/project1/main.tf
/project2/main.tf
```

When you do a `terraform init` in both these project directories it will result in two `.terraform/plugin` directories inside those project directories:

```
/project1/.terraform/plugins
/project2/.terraform/plugins
```

In Terraform Cloud you will probably have two workspaces that both connect to the same repo and have either `project1` or `project2` defined as working directory. When you now do a `terraform plan` in one of these project directories Terraform will try to find the root directory of the repo (`/` in this case).

That root path will then be passed to the `Pack` function in this package and will be used to create and upload the slug. So I don't see how your proposed changes would filter out (for example) the `.terraform` directory in `/project2` when running a `terraform plan` from `/project1`?

Instead it looks like it would strip out the `.terraform` directories of both project directories which would potentially break a Terraform run inside Terraform Cloud. Do you agree on that or am I missing anything?

0gajun:
Thanks for your kind review and comment :)

> In Terraform Cloud you will probably have two workspaces that both connect to the same repo and have either `project1` or `project2` defined as a working directory. 

Yes, that's right.

> When you now do a `terraform plan` in one of these project directories Terraform will try to find the root directory of the repo (`/` in this case).
> 
> That root path will then be passed to the `Pack` function in this package and will be used to create and upload the slug. So I don't see how your proposed changes would filter out (for example) the `.terraform` directory in `/project2` when running a `terraform plan` from `/project1`?

`src` argument of `Pack` function will be `/` in that case, right?
It'll contain both `projet1` and `project2` directories.
So, I think that `filepath.Walk(src, packWalkFn(...))` (https://github.com/hashicorp/go-slug/pull/5/files#diff-cb0d93a0795bd02ed21c986b58f97f14R40) will walk all files and directories under both project directories, even when running a `terraform plan` in `/project1/`.

Let me show my proposed diff. Current changes are like following.

```diff
        // Skip .terraform subdirectories, except for the modules subdirectory.
-        if strings.HasPrefix(subpath, ".terraform"+string(filepath.Separator)) &&
-            !strings.HasPrefix(subpath, filepath.Clean(".terraform/modules")) {
+        if info.IsDir() &&
+            filepath.Base(filepath.Dir(subpath)) == ".terraform" &&
+            info.Name() != "modules" {
            return filepath.SkipDir
        }
```

With the previous solution, to skip `.terraform` subdirectories, except for the `module` subdirectory, `subpath` in the code should start with `.terraform/` due to using `strings.Hasprefix` function.

However, with the multiple project (like this PR's example), `subpath` starts with `project1/` or `project2/`. So, `packWalkFn` cannot return `filepath.SkipDir` when `subpath` is `project1/.terraform/plugins` or `project2/.terraform/plugins`.

My proposed change will resolve this issue.
When `subpath` is `project1/.terraform/plugins`, `filepath.Dir(subpath)` returns `project1/.terraform`, and `filepath.Base(filepath.Dir(subpath))` returns `.terraform`. So, `packWalkFn` can return `filepath.SkipDir` correctly, and `Pack` function doesn't include the `project1/.terraform/plugins` subdirecotry.

While, when `subpath` is `project1/.terraform/modules`, `info.Name()` is `modules`. So, `packWalkFn` doesn't return `filepath.SkipDir` and `Pack` function includes the `project1/.terraform/modules` subdirectory correctly.

Does it make sense...?

> Instead it looks like it would strip out the .terraform directories of both project directories which would potentially break a Terraform run inside Terraform Cloud. Do you agree on that or am I missing anything?

I think this change doesn't break it.
Original implementation has stripped out the `.terraform` subdirectories, except the `module` subdirectory. 
My implementation also ignore `.terraform` subdirectories, but not the `module` subdirectory. I think these are the same behavior for the projects.

Please let me know if I have misunderstood.

svanharmelen:
Hi @0gajun, thank you so much for this PR as it made me realize we needed to fix something 😏 Please have a look at PR #7 and maybe even (if possible) use it yourself a bit.

I know it introduces a breaking change, but I think we should be OK as this slug package is pretty much only used by `go-tfe` which we can update after releasing a new version of this package.

Again thanks and I hope you are fine with me superseding this PR by a slightly bigger PR!

0gajun:
Sorry for the late reply.
Thanks for implementing my idea differently.
I never thought there is a choice making breaking changes 👍 

I'll look and try your PR this weekend (I'm on vacation 😎 ). I'll close this PR.

0gajun:
Implemented in #8

PR #7, opened 2019-09-08
Be more specific about what to include in the slug

svanharmelen:
Especially for slugs created by Terraform i.c.w. configuring a working
directory or when multiple workspaces are configured to use the same
repo.

This work was inspired by PR #5 created by @0gajun. While that PR
improves things a bit already, you could still end up with pretty big
slugs depending on your setup.

We have a specific repo that is used by 5 workspaces each with their
own working directory. Without this PR the slug is about ~490MB, with
these changes the slug is just ~15MB!


svanharmelen:
FYI: This PR does introduce a breaking change but as far as I can tell this package is only used by `go-tfe`, so it should be pretty straight forward to update `go-tfe` and with that `Terraform` and the TFE provider.

I will create the required PR's for that after we agreed on this PR so that we can all update them in sync.

svanharmelen:
@ryanuber it would be nice if we can move this forward. So if you have a minute 😏

slug.go
```diff
@@ -75,19 +75,27 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
 			return nil
 		}
 
-		// Ignore the .terraform directory itself.
-		if info.IsDir() && info.Name() == ".terraform" {
+		// Ignore the .terraform and terraform.d directories.
+		if info.IsDir() && info.Name() == ".terraform" || info.Name() == "terraform.d" {
 			return nil
```
mildwonkey: Is it possible to `return filepath.Skipdir` earlier and skip some of the logic below, or am I missing something?

svanharmelen: We cannot return `filepath.Skipdir` here because we want to include `.terraform/modules`

mildwonkey: 🤦‍♀ yes thank you, that makes perfect sense, sorry!

svanharmelen: No problem at all! 👍

ryanuber:
@svanharmelen thanks! I think there are two different use cases in play here:

**Terraform OSS**
This change makes total sense for the CLI side of Terraform. There is a clear workspace in context, and a clear set of configuration that must be uploaded, so excluding many potentially large directories which will never be needed from other parts of the working directory makes sense.

**Terraform Cloud**
In this case, which you mentioned in the description of this PR, you may have many workspaces hooked up to the same VCS repository, which is valid. However, I think you are assuming that `Pack()` is called for each workspace individually. In the case of Terraform Cloud, `Pack()` is called just once for each commit. If multiple workspaces in the same organization use the same repository, each workspace is associated with the same new `ConfigurationVersion`.

Given the above, Terraform Cloud wouldn't be able to leverage this same approach. It does work for OSS so I think this could still be valid, but I wanted to make sure that this was your only target use case, to avoid setting the expectation that this fixes any repository size/performance related issues in Terraform Cloud when using monorepos.

I'll read the code more closely tomorrow but wanted to at least get back to you with my initial thoughts.

svanharmelen:
> In the case of Terraform Cloud, Pack() is called just once for each commit

I know, but I assumed that Terraform Cloud will call `Pack()` without a specific prefix (so passing an empty string as the prefix) which should then result in the same slug as before.

At least that was the intention 😏 While double checking the code just now I see that it's probably currently not the case. So, assuming you still intent to use this package internally as well, I will update the PR to fix the behavior.

ryanuber:
@svanharmelen I thought about this a bit more today, and realized that there are some workflows that will be broken by this change.

If you perform a remote `terraform plan`, the slug will be uploaded preserving only a specific prefix. In Terraform Cloud, it's possible to use the "Queue Plan" button to queue new plans re-using the last uploaded ConfigurationVersion. It's also possible to modify the workspace settings to look at alternate working directories. From this point of view, it seems that the use of the server-side working directory setting and the proposed functionality here are at odds.

#5 had a more generic solution which would simply expand the file stripping to include all `*/.terraform/*` files, minus any modules in `*/.terraform/modules/*`. I realize that this does not include removal of `terraform.d` directories, though.

I feel like #5 would have more predictable results (even if the working directory setting changes). That seems like a safer change to me at the moment.

What do you think?

svanharmelen:
> In Terraform Cloud, it's possible to use the "Queue Plan" button to queue new plans re-using the last uploaded ConfigurationVersion

Are you sure about that? Does that include configuration versions that are uploaded using `speculative = true`? I very much doubt that and think configuration versions using `speculative = true` are always used only for that specific run. In which case this would not cause any issues when using `Queue Plan` of when changing working directory settings.

> #5 had a more generic solution which would simply expand the file stripping to include all `*/.terraform/*` files, minus any modules in `*/.terraform/modules/*`

That is true, but for repos that are connected to lots of workspaces and so also have lots of different Terraform working directories configured, this means that the uploaded slug is still very considerable in size. Using the approach in this PR really makes a world of difference in usability (with regards to the time needed to create and upload the slug).

So I still would like to implement this solution so it can be used when uploading a speculative plan. Does that make sense? Maybe we can have a small chat on Slack (you can still find me there) if that would make it easier/quicker to find a way forward. I do understand that I have to make the logic work in both cases through. Have some ideas about that, so will see if I can update the PR later today.

ryanuber:
@svanharmelen yes, configuration versions may be reused. You're right that in the case of a _speculative_ plan, the configuration version isn't likely going to be reused in any real capacity. However, if you run a remote apply, the configuration version would indeed be used by a subsequent "Queue Plan", which would potentially be impacted by a slug trimmed specifically for a given working directory.

#5 should strip down slug sizes considerably, removing all of the plugins and other large-format files from the entire slug. I'm open to further ways we could optimize away the extraneous `terraform.d/` directories but I do think we need to ensure we don't introduce new corner cases which impact other users.

Perhaps we could apply the logic in this PR only to speculative plans in the CLI? This way we can be sure that the slug isn't really usable any place else. A `terraform apply` with remote operations would still end up uploading the slug including the `terraform.d` directories (but trimming out other binaries and such more aggressively). That would help preserve all functionality and at the same time help speed up the dev/test loop considerably. I'm not sure if this is a useful compromise, though.

There is also a separate effort just kicking off which should allow `.gitignore`-style file selection for Terraform Cloud projects. This will be a huge improvement, and it can avoid many of the issues I've outlined here. I am hoping something more vanilla like #5 would be enough to tide us over.

ryanuber:
FYI #8 is introducing the ability to exclude files explicitly and changing a few defaults. This should completely cover the use cases from #5.

As for this PR, it does still allow stripping down the slug for a specific workspace, however this should probably be revisited once #8 is finalized and merged to see if it still makes sense to approach the problem in this way.

svanharmelen:
Yeah I was already looking at PR #8 and was thinking the same. Maybe we should close this PR in favor of PR #8 for now and indeed revise how this could/should work once that PR has landed.

I still think it makes sense to add something that improves running `plan` locally, but I agree it need a little more thought. And unfortunately I've been super busy the last weeks so didn't have to come back to this PR yet.

svanharmelen:
Closing this one for now. Maybe I'll take another look (when time permits) on how to add something like this using the updated ignore approach.

PR #8, opened 2019-10-09
Add support for .terraformignore feature

pselle:
Migrates existing exclusions (.git directory and .terraform excluding .terraform/modules) to more global usage (all .git directories and all .terraform directories [exclusive of modules], versus only those at root level).

Adds support for a ".terraformignore" file that may be located at the configuration root to define files that should be included or excluded from the slug.

Before merge/remaining work:

- [x] Move debugging "skipping" lines behind environment flag
- [x] Update Terraform CLI message for when using working dir setting in TFC
- [x] Docs in terraform-website https://github.com/hashicorp/terraform-website/pull/960
- [x] Docs in /terraform repo for remote backend


terraformignore.go
```diff
@@ -0,0 +1,242 @@
...
+	r.regex = re
+	return nil
+}
+
+/*
+	Default rules:
+	.git/
+	.terraform/
+	!.terraform/modules/
+*/
```
svanharmelen: Shouldn't this be the other way around?

pselle:
@svanharmelen Unawareness/not finding this implementation! Thank you for the invitation, I'll look at this and might take you up on the offer.

terraformignore.go
```diff
@@ -0,0 +1,219 @@
...
+		// Only show the error debug if an error *other* than IsNotExist
+		if !os.IsNotExist(err) {
+			fmt.Printf("Error reading .terraformignore, default exclusions will apply: %v \n", err)
+		}
+		return defaultExclusions
+	}
+	return readRules(file)
+}
+
+func readRules(input io.Reader) []rule {
```
jbardin: Not saying we need to this at the moment, but this might be more efficient if the ignore rules are compiled into single state machine rather than a series of rules to iterate over for each path. It would really only be useful for very large numbers of files, which terraform configurations usually don't contain, but might be an easy win for those exceptional situations.

slug.go
```diff
@@ -75,20 +74,16 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
...
 
-		// Skip .terraform subdirectories, except for the modules subdirectory.
-		if strings.HasPrefix(subpath, ".terraform"+string(filepath.Separator)) &&
-			!strings.HasPrefix(subpath, filepath.Clean(".terraform/modules")) {
-			return filepath.SkipDir
+		// Catch directories so we don't end up with empty directories,
+		// the files are ignored correctly
+		if info.IsDir() {
+			if m := matchIgnoreRule(subpath+string(os.PathSeparator), ignoreRules); m {
+				return nil
```
jbardin: I think this should still return `filepath.SkipDir`. The end result is the same, but prevents having to match and exclude every single file recursively in excluded directories. That would be a big difference for directories with a very large number of entries, like `.git/`.

pselle: The problem/issue with using `SkipDir` is that a child of the ignored directory might be included :/

jbardin: Ah, I wasn't thinking about exceptions to allow paths within a skipped directory. Yeah, in that case I guess we have to check all files regardless.

terraformignore.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, rule := range rules {
+		match, _ := rule.match(path)
+
+		if match {
+			matched = !rule.excluded
+		}
+	}
+
+	if matched {
+		fmt.Printf("Skipping excluded path: %s \n", path)
```
jbardin: Should this also use the `debug` call?

terraformignore.go
```diff
@@ -0,0 +1,219 @@
...
+func parseIgnoreFile(rootPath string) []rule {
+	// Look for .terraformignore at our root path/src
+	file, err := os.Open(filepath.Join(rootPath, ".terraformignore"))
+	defer file.Close()
+
+	// If there's any kind of file error, punt and use the default ignore patterns
+	if err != nil {
+		// Only show the error debug if an error *other* than IsNotExist
+		if !os.IsNotExist(err) {
+			fmt.Printf("Error reading .terraformignore, default exclusions will apply: %v \n", err)
```
jbardin: I think errors should be going to stderr in general, but I'm not sure where the output of this fits into the Terraform UI.

svanharmelen: I very much doubt if a package like this should write anything to either `stdout` or `stderr`. Wouldn't it be cleaner to return any errors or warnings and let the caller decide what to do with them?

pselle: Good point, I'll update it at least to go to standard error.

I'm also a bit torn on what to do about Terraform UI -- since this is consumed by go-tfe, the function in go-tfe would need to update to handle errors, and I was attempting to avoid needing to change the function signature in this round, but if _that_ handles the error, it can then pass that up again through to the remote backend, at which point we have tfdiags at our disposal to show things the "right" way.

terraformignore.go
pselle: It will eventually be behind a flag, yes, while people are testing I have it noisy so people can see what's happening.

terraformignore.go
pselle: Probably! In research I saw some implementations that stored rules in different data structures to make match-checking faster, so it's a fair point to make, but I think this is fast enough for now.

terraformignore.go
pselle: Practically, yes, that would be correct. However, the caller in this case will be the `Pack` function, which would then have this same decision, and upstream go-tfe currently bails on any error case.

0gajun:
Great feature! Thanks a lot!

PR #13, opened 2020-12-04
enable GitHub code scanning workflow

chair6:
We're in the process of adopting GitHub "Advanced Security", which includes CodeQL code scanning capabilities.

This sets up the code scanning workflow to run against master as a scheduled weekly job & also on-demand via Actions UI. Results of scans will be available only to users with 'write' access to this repo, via the 'Security' tab (https://github.com/hashicorp/go-slug/security/code-scanning).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,10 +1,9 @@
-name: "Code scanning - action"
+name: "Code scanning - scheduled (weekly) or on-demand"
 
 on:
-  push:
-    branches: [ enable-gh-code-scanning ]
-  pull_request:
-    branches: [ enable-gh-code-scanning ]
+  schedule:
+    - cron: '0 15 * * 0'
+  workflow_dispatch:
 
 jobs:
   CodeQL-Build:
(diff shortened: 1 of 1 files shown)


eastebry (approved):
👍

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,50 @@
...
+      if: ${{ github.event_name == 'pull_request' }}
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      # Override language selection by uncommenting this and choosing your languages
+      with:
+        languages: go
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
```
RussellRollins, line 33: It seems like either auto or manual build should be enabled, if I'm reading these comments correctly.

But also it seems to be working without this, so 🤷 .

chair6: Yup, it's working - code scanning doesn't need auto/manual build for Go projects.  Those comments are boilerplate from the GitHub template, which I'll remove/update as we build out our own version...

PR #14, opened 2020-12-07
Classify errors related to illegal input as such.

RussellRollins:
This PR:

1. Upgrades Go to 1.15 for new [Go 1.13 error behavior](https://blog.golang.org/go1.13-errors).
1. Adds a simple CircleCI config to run tests on all PRs.
1. [Removes leading capitalization](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) from all error strings. 
1. Switches errors related to illegal paths and symlinks to a new error type `IllegalSlugError`


ryanuber:
I agree with the capitalization change but it makes me worry a little about importers who might be error checking. I think we probably still proceed anyways, just something to be aware of and consider.

slug.go
```diff
@@ -19,6 +19,23 @@ type Meta struct {
 	Size int64
 }
 
+// IllegalSlugError indicates the provided slug (io.Writer for Pack, io.Reader
+// for Unpack) validates a rule about its contents. For example, an absolute or
```
ryanuber, line 23:
```suggestion
// for Unpack) violates a rule about its contents. For example, an absolute or
```

slug.go
```diff
@@ -19,6 +19,23 @@ type Meta struct {
...
+// external symlink.
+type IllegalSlugError struct {
+	Err error
+}
+
+// Error implements the error interface by returning an error string about the
+// issue with the underlying error.
```
ryanuber, line 30: I would usually not document the implementation of a known interface like this. It's a hard sentence to read and I think stating that `IllegalSlugError` implements the `error` interface on the struct is sufficient.

ryanuber (approved):
LGTM - just minor feedback.

RussellRollins:
@ryanuber any opinion about what version to release this under? I'm tempted to call it `0.5.1`, since the changes only affect you if you have been keying off of error text. Plus it's a sub 1.0 release, so I generally feel okay playing fast and loose.

But also _we_ have code that keys off of error text, so this would actually cause (very minor) regressions in a few spots if it's updated before we fix them.

PR #16, opened 2021-02-24
Allow bypassing the .terraformignore rules

RussellRollins:
While Terraform packing up configuration is one use of go-slug's Pack function, there are other uses not directly related to Terraform.

To that end, it'd be nice for the `.terraformignore` functionality to be optional. However, at the moment, it is not.

Even if `go-slug` user intentionally creates a blank `.terraformignore` file, the system is still "additive". The default rules are always applied.

This PR introduces a new struct, the `Packer` struct (ironically, completely unrelated to Packer the HashiCorp product).

The existing package level Pack function will keep its same function signature, but callers of `go-slug` can now create a custom Packer using functional options, which will have a simplified function signature.

One of those options will be to bypass the `.terraformignore` rules in this case, all the files in the src directory will be packed, no matter what.

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -77,6 +77,14 @@ func NewPacker(options ...PackerOption) (*Packer, error) {
 	return p, nil
 }
 
+// Pack at the package level is used to maintain compatibility with existing
+// code that relies on this function signature. New options related to packing
+// slugs should be added to the Packer struct instead.
+func Pack(src string, w io.Writer, dereference bool) (*Meta, error) {
+	p := Packer{dereference: dereference}
+	return p.Pack(src, w)
+}
+
 // Pack creates a slug from a src directory, and writes the new slug
 // to w. Returns metadata about the slug and any errors.
 //
--- a/slug.go
+++ b/slug.go
@@ -37,11 +37,11 @@ func (e *IllegalSlugError) Unwrap() error { return e.Err }
 // PackerOption is a functional option that can configure non-default Packers.
 type PackerOption func(*Packer) error
 
-// BypassIgnore is a PackerOption that will skip the .terraformignore rules and
-// Pack the entire src directory.
-func BypassIgnore() PackerOption {
+// ApplyIgnore is a PackerOption that will apply the .terraformignore rules and
+// skip packing files it specifies.
+func ApplyIgnore() PackerOption {
 	return func(p *Packer) error {
-		p.bypassIgnore = true
+		p.applyIgnore = true
 		return nil
 	}
 }
@@ -57,15 +57,15 @@ func DereferenceSymlinks() PackerOption {
 
 // Packer holds options for the Pack function.
 type Packer struct {
-	dereference  bool
-	bypassIgnore bool
+	dereference bool
+	applyIgnore bool
 }
 
 // NewPacker is a constructor for Packer.
 func NewPacker(options ...PackerOption) (*Packer, error) {
 	p := &Packer{
-		dereference:  false,
-		bypassIgnore: false,
+		dereference: false,
+		applyIgnore: false,
 	}
 
 	for _, opt := range options {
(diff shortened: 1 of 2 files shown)


pselle (commented):
I'm a bit confused how this is passing tests when other tests aren't updated to use `Packer` instead of only `Pack` 😬

RussellRollins:
@pselle the package level Pack function still exists:

https://github.com/hashicorp/go-slug/pull/16/files#diff-441cb02ae7533f00597151444867785b1c71db70d6118d54c3e8a577cd2e21d4R83

It's just a wrapper around creating a Packer with the input dereference value, and calling Pack.

This should be an invisible change for callers using slug.Pack, if I did things right.

slug.go
```diff
@@ -34,14 +34,65 @@ func (e *IllegalSlugError) Error() string {
 // chain.
 func (e *IllegalSlugError) Unwrap() error { return e.Err }
 
+// PackerOption is a functional option that can configure non-default Packers.
+type PackerOption func(*Packer) error
+
+// BypassIgnore is a PackerOption that will skip the .terraformignore rules and
+// Pack the entire src directory.
+func BypassIgnore() PackerOption {
```
ryanuber, line 42: Instead of having an option to _disable_ the `.terraformignore` file, we should consider having an option to _enable_ it. This package is used in many contexts that are not Terraform, an as such I think that the default should be as unsurprising as possible.

We can still maintain the package-level `Pack()` and have that default to applying the `.tfignore`, such that we are fully backwards compatible and do not introduce any regressions. But I'd strive for `NewPacker().Pack()` to default to packaging in a generic way without any fancy .terraformignore logic.

RussellRollins: This makes a lot of sense to me. It fits well with the "make the zero value useful" rule, and it fixes my awkward "if not not applying" if statement. And like you said, since we can set whatever rules in the package level `Pack` we want, it can be invisible to current callers. Will do.

slug.go
```diff
@@ -34,14 +34,71 @@ func (e *IllegalSlugError) Error() string {
 // chain.
 func (e *IllegalSlugError) Unwrap() error { return e.Err }
 
+// PackerOption is a functional option that can configure non-default Packers.
+type PackerOption func(*Packer) error
+
+// ApplyIgnore is a PackerOption that will apply the .terraformignore rules and
+// skip packing files it specifies.
+func ApplyIgnore() PackerOption {
```
ryanuber, line 42: Nitpick: I'd kind of like this better as `ApplyTerraformIgnore`. It's more explicit and obvious what it does.

slug_test.go
```diff
@@ -240,6 +240,79 @@ func TestPackWithoutDereferencing(t *testing.T) {
...
+		}
+	}
+	if !bazFound {
+		t.Fatal("expected file baz.txt to be present, but not found")
+	}
+
+	// .terraform/file.txt would normally be ignored, but should not be
+	var dotTerraformFileFound bool
+	for _, file := range fileList {
+		if file == ".terraform/file.txt" {
```
ryanuber, line 298: It might be even more clear to write a `.terraformignore` file which explicitly ignores `baz.txt` or something. This works fine, but it relies on the current defaults.

ryanuber (approved):
LGTM! Just very minor comments.

slug_test.go
RussellRollins: This test slugs up `testdata/archive-dir`, which does have an explicit `.terraformignore` that ignores `baz.txt`. The test that that file is not skipped is above this one, so this is testing the default rules as well as a rule created by `testdata/archive-dir/.terraformignore`.

We could write the `.terraformignore` file inline in this test, but that doesn't really seem better than using the `testdata` directory to me.

ryanuber: Ah, I missed that the .terraformignore was already in the slug. I think that's good. Thanks for clarifying!

pselle (approved):
Looks good!

Issue #20, opened 2021-07-21
Avoid re-compiling rules


skeggse:
I observed, while looking into #19, that due to the current implementation we're actually re-compiling every rule for every file. Updating the `range` in `matchIgnoreRule` to be by-reference improves performance on large repositories by a factor of ~3.

apparentlymart:
I independently noticed the same thing while working on ccbbeeefcd661b65b9b2141ca5254c38bae49fe6 and attempted to improve it, but since performace wasn't my main motivation in that commit I didn't spend much time studying the impact of the change I attempted there.

Looking back at it again with fresh eyes today I suspect I may not have actually fully addressed it: @skeggse's note about passing by reference instead of by value makes me notice that I seem to have made it just ineffectually precompile _copies_ of the default rules, rather than the actual values that end up in the default ruleset:

https://github.com/hashicorp/go-slug/blob/6a7ca7ac026f4d9fd54540d46c61e35caf4e33e5/internal/ignorefiles/terraformignore.go#L161-L186

A small change that might make it effectual is to refer directly to the array elements by index, rather than working on the copies that the `for ... range` is implicitly making:

```hcl
	for i := range defaultExclusions {
		err := defaultExclusions[i].compile()
		if err != nil {
			panic(fmt.Sprintf("invalid default rule %q: %s", r.val, err))
		}
	}
```

The default match object includes a slice value covering the same backing array, so I _expect_ that a change like the above would make it effective, but I've not yet tested.

skeggse:
@brandonc i don't suppose you looked at this issue while you were working on the directory issue?

PR #21, opened 2021-07-21
perf: improve pack performance, implement tree pruning

### LINKED ISSUE #19: Avoid entering excluded directories
Let's say we're in a large monorepo, and we'd like to exclude all the `node_modules/` directories (they might not be checked in, even). It'd be nice if I could put `node_modules/` in the `.terraformignore` file and have it not even traverse that directory - it seems that's how it currently works, because the total runtime for `Pack` seems to remain the same even with that directive.

skeggse:
On a large repository (2^17 files, mostly ignored by git), this plus an appropriate `.terraformignore` improves performance from ~60 seconds to ~3 seconds.

Fixes #19 and #20.

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -149,14 +149,21 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
 			return nil
 		}
 
-		if m := matchIgnoreRule(subpath, ignoreRules); m {
-			return nil
+		if m, r := matchIgnoreRule(subpath, ignoreRules); m && !(r.val == "**/*" && info.IsDir()) {
+			if info.IsDir() {
+				return filepath.SkipDir
+			} else {
+				return nil
+			}
 		}
 
 		// Catch directories so we don't end up with empty directories,
 		// the files are ignored correctly
 		if info.IsDir() {
-			if m := matchIgnoreRule(subpath+string(os.PathSeparator), ignoreRules); m {
+			if m, r := matchIgnoreRule(subpath+string(os.PathSeparator), ignoreRules); m {
+				if r.val != "**/*" { // Represents the `*` rule
+					return filepath.SkipDir
+				}
 				return nil
 			}
 		}
--- a/terraformignore.go
+++ b/terraformignore.go
@@ -74,22 +74,28 @@ func readRules(input io.Reader) []rule {
 	return rules
 }
 
-func matchIgnoreRule(path string, rules []rule) bool {
-	matched := false
+func matchIgnoreRule(path string, rules []rule) (bool, *rule) {
+	var matched *rule = nil
 	path = filepath.FromSlash(path)
-	for _, rule := range rules {
+	for idx := range rules {
+		rule := &rules[idx]
 		match, _ := rule.match(path)
 
 		if match {
-			matched = !rule.excluded
+			if rule.excluded {
+				return false, rule
+			}
+			if rule.val != "**/*" || matched == nil {
+				matched = rule
+			}
 		}
 	}
 
-	if matched {
+	if matched != nil {
 		debug(true, path, "Skipping excluded path:", path)
 	}
 
-	return matched
+	return matched != nil, matched
 }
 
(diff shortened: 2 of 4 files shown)


slug.go
```diff
@@ -149,14 +149,21 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
 			return nil
 		}
 
-		if m := matchIgnoreRule(subpath, ignoreRules); m {
-			return nil
+		if m, r := matchIgnoreRule(subpath, ignoreRules); m && !(r.val == "**/*" && info.IsDir()) {
```
skeggse, line 152: I'm not terribly happy about having to include `**/*` in here, but it makes it possible to write ignorefiles like:
```ignore
.git/
node_modules/
*
!*.tf
!*.json
!*.hcl
```
and get only the `tf`, `json` and `hcl` files. Without this, the `*` matches all directories and prevents recursion into anything that isn't explicitly allowed.

terraformignore.go
```diff
@@ -74,22 +74,28 @@ func readRules(input io.Reader) []rule {
...
 	path = filepath.FromSlash(path)
-	for _, rule := range rules {
+	for idx := range rules {
+		rule := &rules[idx]
 		match, _ := rule.match(path)
 
 		if match {
-			matched = !rule.excluded
+			if rule.excluded {
+				return false, rule
```
skeggse, line 86: This precedence matches `git`'s behavior, in that `git` will see `!`-prefixed lines as overriding other inclusive rules.

terraformignore.go
```diff
@@ -200,9 +206,13 @@ var defaultExclusions = []rule{
 		excluded: false,
 	},
 	{
-		val:      filepath.Join("**", ".terraform", "**"),
+		val:      filepath.Join("**", ".terraform", "?**"),
```
skeggse, line 209: In order to actually discovery `.terraform/modules`, we have to explicitly allow recursion into `.terraform` but not anything under it.

terraformignore.go
```diff
@@ -74,22 +74,28 @@ func readRules(input io.Reader) []rule {
...
 }
 
-func matchIgnoreRule(path string, rules []rule) bool {
-	matched := false
+func matchIgnoreRule(path string, rules []rule) (bool, *rule) {
+	var matched *rule = nil
 	path = filepath.FromSlash(path)
-	for _, rule := range rules {
+	for idx := range rules {
+		rule := &rules[idx]
```
skeggse, line 81: i don't know if there's an idiomatic go way to express this, but this is a pretty significant performance improvement

felipeplets:
@brandonc is there anything concerning you in this PR?

brandonc:
@felipeplets The package has changed quite a bit since this PR was developed so it needs to be revisited. Aside from that, my main concern with this PR is the seemingly magic rule `**/*`

Imagine trying to prevent recursion within a large directory:

```
src/**/*
!src/this/one/folder
```

This would be exempt from the recursion skip and revert back to being very slow.

This suggestion does make me wonder if the rules could be defined by the parser such that the directory walk could be short-circuited if there are no exclusions defined that would pertain to an ignored directory.

brandonc:
@skeggse @felipeplets I merged #50 (but have not yet integrated it with Terraform). It addressed the problem a little more narrowly but avoids special cases. The bottom line is this: In the future, if you must use negation rules in your .terraformignore file, place them as high as possible on the list of rules. I was able to optimize for the situation where a directory can be fully ignored if no negation rules follow. Consider:

```
node_modules/
logs/**/*
!logs/production/**/*
```

With the negated rule present, .terraformignore cannot rule out the entire `node_modules` directory because there might be a "logs" directory inside there somewhere. But, since rules are designed to be allowed to override previous rules, if you move that negation above node_modules, it can now be safely ignored.

```
logs/**/*
!logs/production/**/*
node_modules/
```

Now .terraformignore can avoid descending into node_modules since there are no rules that could possibly include files that follow it.

PR #22, opened 2021-12-18
update references to default branch

radditude:
This PR updates references to the default branch as part of the effort to rename the default branch to `main` across our repos.

The CI failure appears to be related to Circle <> Github permissions - it looks like we'll need to recreate the Circle SSH key.


brandonc:
This repo was migrated to GHA so maybe the CirclCI issue has disappeared!

PR #24, opened 2022-02-25
Add terraform.tfstate to default exclusion list

barrettclark:
TFE/TFC will return an error if `terraform.tfstate` is detected in the
slug that is uploaded to the remote agent. In TFE/TFC we read state from
the cloud rather than a state file, so we do not need `terraform.tfstate`
to be included in the slug.

This commit adds `terraform.tfstate` to the default exclusions list,
which is used if, and only if, there is no `.terraformignore` file
specified in the project directory. So if that file is included the
`terraform.tfstate` file WILL be included in the slug, unless of course
it is added to the `.terraformignore` file.

This gives us flexibility to have the default state file included if
needed and excluded by default.

### DIFF
--- a/terraformignore.go
+++ b/terraformignore.go
@@ -192,6 +192,7 @@ func (r *rule) compile() error {
 	.git/
 	.terraform/
 	!.terraform/modules/
+	terraform.tfstate
 */
 
 var defaultExclusions = []rule{
@@ -206,6 +207,10 @@ var defaultExclusions = []rule{
 	{
 		val:      filepath.Join("**", ".terraform", "modules", "**"),
 		excluded: true,
+	},
+	{
+		val:      filepath.Join("**", "terraform.tfstate"),
+		excluded: false,
 	},
 }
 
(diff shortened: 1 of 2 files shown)


ryanuber (approved):
This makes sense to me. It'd be good to get @chrisarcand to weigh in maybe, in case there are any cases I'm not thinking about that this would break.

nfagerlund (approved):
For what it's worth, this makes perfect sense to me too. I can't think of any case where you'd want to upload a terraform.tfstate file (in any subdirectory) as part of a TFC config slug.

barrettclark:
@ryanuber What do I need to do to clear the CircleCI failure? It looks like it didn't even make it to running the tests.

ryanuber:
@barrettclark my guess is that the SSH key used to clone was revoked at some point. We might have used a GitHub deploy key. Currrently go-slug is using the deploy key with fingerprint `d2:81:ec:32:b9:dc:13:d9:be:bc:ae:95:53:49:25:e1`. It appears that I don't have admin access on this repo anymore, so I can't verify if that's correct or if that got pulled at some point.

chrisarcand (commented):
Sorry for the delay!

I think this is great, and maybe even _too_ cautious 😄 Given that any runner can't operate at all, it seems reasonable to always omit it, though I appreciate the level of caution. 

But this is a very reasonable compromise, catering to the primary persona we wanted to fix this for - newly migrated users who likely haven't defined at sort of ignore file. 👍🏻

PR #27, opened 2022-06-17
Support dangling symlinks

### LINKED ISSUE #18: Terraform fails to build a slug when there is a symlink leading to a non-existing directory
See https://github.com/hashicorp/terraform/issues/28109

ryanuber:
Fixes an issue where a dangling symlink causes the `Pack()` function to fail. In general a dangling symlink is fairly innocuous, particularly so if the symlink is pointing to a path within the "root" of the archive. With this PR , dangling symlinks are now allowed so long as their targets remain _within_ the root of the archive.

This also adds some additional tests to verify that we handle dangling symlinks on the `Unpack()` side as well, and disallows absolute symlinks from being packed into a slug, which would inevitably error on the `Unpack()` side.

Fixes #18

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -193,10 +193,24 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
 			header.Size = info.Size()
 
 		case fm&os.ModeSymlink != 0:
-			target, err := filepath.EvalSymlinks(path)
+			// First read the symlink file to find the destination.
+			target, err := os.Readlink(path)
 			if err != nil {
 				return fmt.Errorf("failed to get symbolic link destination for %q: %w", path, err)
 			}
+
+			// Disallow creating symlinks with absolute paths.
+			if filepath.IsAbs(target) {
+				return &IllegalSlugError{
+					Err: fmt.Errorf(
+						"invalid symlink (%q -> %q) has absolute target",
+						path, target,
+					),
+				}
+			}
+
+			// Get the path to the target relative to path.
+			target = filepath.Join(filepath.Dir(path), target)
 
 			// If the target is within the current source, we
 			// create the symlink using a relative path.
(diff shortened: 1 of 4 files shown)


slug.go
```diff
@@ -193,11 +193,25 @@ func packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta, dereference
...
 
 		case fm&os.ModeSymlink != 0:
-			target, err := filepath.EvalSymlinks(path)
+			// First read the symlink file to find the destination.
+			target, err := os.Readlink(path)
 			if err != nil {
 				return fmt.Errorf("failed to get symbolic link destination for %q: %w", path, err)
 			}
 
+			// Disallow creating symlinks with absolute paths.
```
gavriel-hc, line 202: Since this is a public repo, and this is maybe a breaking change, should we bump the version?

gavriel-hc (approved):

slug.go
ryanuber: I do think we should bump the version. We are still in 0.x on this repo, which gives us some space to make important fixes like this, so I don't think we should for example bump to 1.0 from 0.x to express the breaking change. Going to a new minor within 0.x is maybe reasonable though.

One further thought is that I was looking at this as more of a bug fix, given that if you `Pack()` an archive with an absolute symlink, `Unpack()` will already refuse to unpack it and throw an error. This just moves that validation even earlier. Of course consumers of this package could be using it for only the `Pack()` function and not really looking at it the same way, so in principle I agree with you.

gavriel-hc: Yeah, that's why I said _maybe_ a breaking change. Since it sounds like the only thing it's "breaking" is something that was already broken anyway, agreed that a major bump definitely does not make sense. Even a patch version may be reasonable, but I'd probably lean toward minor

eastebry (approved):
Yup, this seems fine to me! 

Looking back through this code, I'm glad we've got such great tests for the unpack method!
https://github.com/hashicorp/go-slug/blob/master/slug_test.go#L550

PR #28, opened 2022-06-22
Allow absolute symlinks within the source dir during Pack()

ryanuber:
This restores previous behavior of converting absolute symlinks to relative ones when the target falls within the source directory during `Pack()`. Fixes an issue after #27 where we explicitly disallowed any type of absolute target during `Pack()`. Now absolute symlinks are only rejected if they fall outside of the source directory.

Also consolidates symlink testing for `Pack()` which cleans up the testdata dir a bit, reduces boilerplate and adds better coverage of more scenarios.


slug_test.go
```diff
@@ -58,25 +57,12 @@ func TestPack(t *testing.T) {
 			}
 			symFound = true
 		}
-
-		if hdr.Name == "dangling_symlink" {
```
gavriel-hc, line 62: I'm still trying to parse the tests, but is this case still handled by the new ones?

gavriel-hc: ah never mind, I see the `nonexistent` test cases now

slug_test.go
```diff
@@ -327,15 +211,241 @@ func TestPackWithoutIgnoring(t *testing.T) {
...
+		},
+		{
+			desc:           "relative nonexistent internal target without dereferencing",
+			absoluteTarget: false,
+			externalTarget: false,
+			createTarget:   false,
+			dereference:    false,
+			expectTypeflag: tar.TypeSymlink,
+		},
+		{
```
gavriel-hc, line 288: I think you missed the true/true/true/true case

slug_test.go
```diff
@@ -327,15 +211,241 @@ func TestPackWithoutIgnoring(t *testing.T) {
...
+			desc:           "absolute nonexistent external target with dereferencing",
+			absoluteTarget: true,
+			externalTarget: true,
+			createTarget:   false,
+			dereference:    true,
+			expectErr:      "no such file or directory",
+		},
+		{
+			desc:           "relative nonexistent external target with dereferencing",
+			absoluteTarget: true,
```
gavriel-hc, line 322:
```suggestion
			absoluteTarget: false,
```

slug_test.go
```diff
@@ -327,15 +211,241 @@ func TestPackWithoutIgnoring(t *testing.T) {
...
+			desc:           "absolute nonexistent external target without dereferencing",
+			absoluteTarget: true,
+			externalTarget: true,
+			createTarget:   false,
+			dereference:    false,
+			expectErr:      "target outside",
+		},
+		{
+			desc:           "relative nonexistent external target without dereferencing",
+			absoluteTarget: true,
```
gavriel-hc, line 338:
```suggestion
			absoluteTarget: false,
```

gavriel-hc (commented):
~I think the change seems reasonable. I think the tests could be a little simpler to parse by generating them with a loop, e.g.~

```
for i := 0; i < 16; i++ {
  tcase := {
    absoluteTarget: i%2 > 0,
    externalTarget: i%4 > 2,
    createTarget: i%8 > 4,
    dereference: i > 8,
  }
  ...
}

```

Actually as I'm writing this I can imagine it being probably harder to parse. Maybe a better solution would just be to make sure that all the cases are "in binary order"? Right now the order is like

1101, 1100, 0101, 0100, 1001, 1000, ... whereas better would be
0000, 0001, 0010, 0011, ...

Sorry if that was worded weirdly, I'm happy to try to be clearer if it was.

PR #29, opened 2022-07-15
Expose option for explicitly allowing certain symlinks with external targets

ryanuber:
This allows the caller to specify any path(s) which symlinks may point to _outside_ of the slug directory. Sometimes having an absolute symlink to a file elsewhere on the system is necessary, while other times allowing such arbitrary targets on symlinks may pose security concerns. This implementation keeps the default as secure as possible, prohibiting external link targets, but allows callers to specify one or more file paths or prefixes which are allowed. It is then up to the specific application to make the determination as to what is or isn't allowed.

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -47,10 +47,25 @@ func ApplyTerraformIgnore() PackerOption {
 }
 
 // DereferenceSymlinks is a PackerOption that will allow symlinks that
-// reference a target outside of the src directory.
+// reference a target outside of the source directory by copying the link
+// target, turning it into a normal file within the archive.
 func DereferenceSymlinks() PackerOption {
 	return func(p *Packer) error {
 		p.dereference = true
+		return nil
+	}
+}
+
+// AllowSymlinkTarget relaxes safety checks on symlinks with targets matching
+// path. Specifically, absolute symlink targets (e.g. "/foo/bar") and relative
+// targets (e.g. "../foo/bar") which resolve to a path outside of the
+// source/destination directories for pack/unpack operations respectively, may
+// be expressly permitted, whereas they are forbidden by default. Exercise
+// caution when using this option. A symlink matches path if its target
+// resolves to path exactly, or if the target is prefixed by path.
+func AllowSymlinkTarget(path string) PackerOption {
+	return func(p *Packer) error {
+		p.allowSymlinkTargets = append(p.allowSymlinkTargets, path)
 		return nil
 	}
 }
@@ -59,6 +74,7 @@ func DereferenceSymlinks() PackerOption {
 type Packer struct {
 	dereference          bool
 	applyTerraformIgnore bool
+	allowSymlinkTargets  []string
 }
 
 // NewPacker is a constructor for Packer.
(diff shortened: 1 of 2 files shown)


ryanuber:
Keeping this in draft while I make some adjustments.

slug.go
```diff
@@ -47,18 +47,34 @@ func ApplyTerraformIgnore() PackerOption {
...
+// AllowSymlinkTarget relaxes safety checks on symlinks with targets matching
+// path. Specifically, absolute symlink targets (e.g. "/foo/bar") and relative
+// targets (e.g. "../foo/bar") which resolve to a path outside of the
+// source/destination directories for pack/unpack operations respectively, may
+// be expressly permitted, whereas they are forbidden by default. Exercise
+// caution when using this option. A symlink matches path if its target
+// resolves to path exactly, or if the target is prefixed by path.
+func AllowSymlinkTarget(path string) PackerOption {
+	return func(p *Packer) error {
+		p.allowSymlinkTargets = append(p.allowSymlinkTargets, path)
```
tbehling, line 68: Per our chat, let's avoid the ambiguity of partial file path prefixes, by appending a slash if one wasn't provided.

slug_test.go
```diff
@@ -361,6 +361,203 @@ func TestPack_symlinks(t *testing.T) {
...
+			desc:   "relative symlink, prefix match",
+			allow:  "../foo/",
+			target: "../foo/bar/baz",
+		},
+		{
+			desc:   "absolute symlink, non-match",
+			allow:  "/zip",
+			target: "/foo/bar/baz",
+			err:    "has external target",
+		},
```
tbehling, line 396: Per above, ensure that allow "/foo" does not match when target is "/foobar".

slug.go
```diff
@@ -449,6 +432,56 @@ func Unpack(r io.Reader, dst string) error {
...
+	// Get the absolute path of the symlink target.
+	var absTarget string
+	if filepath.IsAbs(target) {
+		absTarget = filepath.Clean(target)
+	} else {
+		absTarget = filepath.Join(filepath.Dir(absPath), target)
+	}
+
+	// Target falls within root.
+	if strings.HasPrefix(absTarget, absRoot) {
```
tbehling, line 460: This might be "belt plus suspenders", but Go has a function [`os.SameFile`](https://pkg.go.dev/os#SameFile), that compares whether two paths point at the same actual inode number (based on `stat`).  That would be one step beyond the string comparison.  However, there'd need to be extra logic if one or the other path doesn't exist.

If we used os.Samefile, there could be something like this pseudocode:

```
statTarget = os.Stat(strings.Left(absTarget, len(absRoot)))
statRoot = os.Stat(absRoot)
if os.SameFile(statRoot, statTarget) {
  // OK
} else {
  // the target is not within the root
}
```

slug.go
ryanuber: This case should be resolved by the latest commit. We can still allow exact matches or specifically directory prefixes.

slug.go
ryanuber: Yeah, we'd definitely need to augment to support the file-doesn't-exist case. We also can't really use this effectively during `Unpack()`, because we'd need to write the file before checking it with `os.SameFile` this way.

tbehling: OK sounds good.  I figured it might be a stretch but just wanted to mention it.

tbehling (approved):
We talked about this and I gave it another read through.  The test cases are great, and cover every combination I can think of offhand.  Nice work!

PR #31, opened 2022-09-22
Set gzip compression level to BestSpeed (1)

brandonc:
If you examine the Persist filesystem span of Agents in TFC, the overwhelming majority of time is spent compressing the filesystem, while a fraction is spent uploading it. This indicates that we should balance more toward less compression at the expense of more data to transmit. DM me for a datadog link.


annawinkler (approved):
This looks OK to me, but I'd really like someone more familiar with the repo to also review.

PR #36, opened 2023-03-15
Resolve root dir if symlink

sebasslash:
This fix results from a bug where if the `src` directory to pack is a symlink, the resulting tar will contain no files. This attempts to resolve the symlink (if necessary) _before_ we begin to walk and archive the contents. 

To quickly reproduce this issue with Terraform:
1. Create any working directory. Ensure that the configuration is using the `cloud` or `backend "remote"` block. Run `terraform init`. 
1. Create a symlink: `ln -s /path/to/wkdir /path/to/wkdir-symlink`
1. `cd /path/to/wkdir-symlink && terraform apply`
1. You should see an error that no configuration files were found.

### DIFF
--- a/slug_test.go
+++ b/slug_test.go
@@ -17,126 +17,33 @@ import (
 
 func TestPack(t *testing.T) {
 	slug := bytes.NewBuffer(nil)
-
 	meta, err := Pack("testdata/archive-dir", slug, true)
 	if err != nil {
 		t.Fatalf("err: %v", err)
 	}
 
-	gzipR, err := gzip.NewReader(slug)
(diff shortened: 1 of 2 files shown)


slug_test.go
```diff
@@ -1101,6 +1008,124 @@ func TestNewPacker(t *testing.T) {
 	}
 }
 
+// This is a reusable assertion for when packing testdata/archive-dir
+func assertArchiveFixture(t *testing.T, slug *bytes.Buffer, got *Meta) {
```
sebasslash, line 1012: Not a huge fan of this naming

brandonc (approved):

PR #39, opened 2023-04-17
Fixed a bug when archives contain entries without a name

SwiftEngineer:
Fixed a bug that occurs when archives contain entries without a name.

### DIFF
--- a/slug_test.go
+++ b/slug_test.go
@@ -1042,7 +1042,10 @@ func TestUnpackEmptyName(t *testing.T) {
 	defer os.RemoveAll(dir)
 
 	// This crashes unless the bug is fixed
-	Unpack(&buf, dir)
+	err = Unpack(&buf, dir)
+	if err != nil {
+		t.Fatalf("err:%v", err)
+	}
 }
 
 // This is a reusable assertion for when packing testdata/archive-dir


slug_test.go
```diff
@@ -1017,6 +1017,34 @@ func TestNewPacker(t *testing.T) {
...
+	}
+
+	dir, err := ioutil.TempDir("", "slug")
+	if err != nil {
+		t.Fatalf("err:%v", err)
+	}
+	defer os.RemoveAll(dir)
+
+	// This crashes unless the bug is fixed
+	Unpack(&buf, dir)
```
ryanuber, line 1045: I'd at least check the returned error here and ensure it is `nil`.

slug.go
```diff
@@ -376,6 +376,12 @@ func (p *Packer) Unpack(r io.Reader, dst string) error {
 
 		// Get rid of absolute paths.
```
ryanuber, line 377: This comment got separated from the thing that actually gets rid of absolute paths - I'd move the comment down to line 385.

ryanuber (commented):
Generally looks good! Super minor comments. 👍

ryanuber (approved):

PR #42, opened 2023-06-05
Source bundler: Unified source code fetching for Terraform CLI and Terraform Cloud

apparentlymart:
So far Terraform Cloud has used an entirely separate mechanism for fetching the source code for a root module than it does for installing any child module packages. This is largely a consequence of how Terraform Cloud started its life as little more than an automation wrapper around Terraform CLI, mimicking how someone would use Terraform CLI locally: a local user typically starts by fetching the source code using something like `git clone`, and then runs `terraform init` to deal with everything else.

As we work to present a more unified experience between these two entry points, it's been increasingly frustrating to have this slightly different treatment in these two cases, since it tends to lead to confusing edge-cases such as how `.terraformignore` files only get handled for the package containing the root module in Terraform Cloud and not in any other case. It also means that Terraform Cloud needs to work pretty hard to preserve the installed child modules between the plan phase and the apply phase since re-running `terraform init` won't necessarily reproduce identical source code, if e.g. new commits have been pushed to the remote Git branch in the meantime.

As a foundation for future Terraform features (not yet planned or announced) that more tightly integrate Terraform CLI and Terraform Cloud, here we implement a carefully-selected subset of Terraform CLI's module installation behavior in a way that can be called from both Terraform CLI and Terraform Cloud, producing an equivalent result in both contexts.

We've added this to `go-slug` because it's already the shared library between Terraform CLI and Terraform Cloud that deals with the problem of packaging up source code. This new functionality introduces the idea of a "source bundle", which is in some sense a "meta-slug": a bunch of source code packages frozen and bundled together into a single directory/archive for later use.

The two new exported packages are explicitly marked as experimental in their documentation because we anticipate needing to make at least some small tweaks to the API as we start to make use of this in some early real-world experiments. We have written a small amount of "realistic-ish" code using this API already to help design it, but need to reserve the possibility of making some breaking changes if we learn that there's something missing. This whole codebase is still currently pre-1.0 anyway, but these extra disclaimers of compatibility mean that the main existing scope of this codebase could potentially reach 1.0 before this new functionality does.

---

Terraform's existing module installer is based on go-getter, which is in retrospect not a great foundation for this sort of problem because it's far too flexible and therefore gives poor user feedback when problems occur and allows many different ways of writing essentially the same source address that make it hard to optimize installation by reusing existing cached packages.

However, we intend to use this only for new features that users would explicitly opt into, and so it's not necessary to fully replicate everything Terraform CLI would allow today: the existing module installer in `terraform init` won't be going anywhere for the foreseeable future, so those who need its more esoteric capabilities can continue using Terraform as they to today.

With that in mind, what we have here is essentially a subset of Terraform CLI's module installation capabilities:

- Installation from Git, whether from a specific commit, a named tag, or another mutable ref.
- Installation from gzipped tar archives fetched over HTTP.
- Indirect installation through Terraform module registries as long as the underlying source location is one of the above. (That includes both the public Terraform Registry and Terraform Cloud's private module registry.)

For everything that this implementation and Terraform CLI have in common, the goal is to use a compatible source address syntax. That means that if someone writes a module for the subset that we support in source bundler then it should _also_ work with Terraform CLI's traditional module installer, but not necessarily vice-versa.

The two initial capabilities described above happen to align exactly with what Terraform Cloud itself can currently fetch for root modules, so it's a natural place to draw the line to start -- we don't need to teach Terraform Cloud about any new installation sources -- and can be expanded later if needed as long as we're adding source types that both Terraform CLI and Terraform Cloud can reasonably support (where Terraform Cloud's multi-tenant nature imposes some additional constraints compared to CLI).

---

Part of the reason to place this new use-case in this existing repository is that this adds support for `.terraformignore` for _all_ source packages -- both for root modules and for downstream modules. The definition of the `.terraformignore` file format belongs to this codebase and is now shared between the single-directory "slug" implementation and the source bundle implementation.

That does mean that sourcebundle-based features will behave _marginally_ differently to Terraform CLI's existing module installer if installing from source packages that have `.terraformignore` files in the root: those files will not be available under sourcebundle-based usage but _will_ be available when installed with Terraform CLI.

This is another example of the source bundle behavior being effectively a subset of the Terraform CLI behavior: it is possible to write a source package that will work in both contexts either by not including `.terraformignore` at all or by making sure that the modules inside don't try to access any of the files that are excluded by `.terraformignore`. Those who just continue using Terraform CLI as-is will be unaffected even if they _do_ have a `.terraformignore` file in their source package.


sourceaddrs/source_registry_final.go
```diff
@@ -0,0 +1,64 @@
+package sourceaddrs
+
+import (
+	"github.com/apparentlymart/go-versions/versions"
```
ryanuber, line 4: Post-hoc comment, but this is the first time I am seeing this - why not hashicorp/go-version? We use that regularly throughout the rest of TFC's codebases.

apparentlymart: We use this library in Terraform because `go-version` tends to generate useless error messages and historically it didn't handle prereleases correctly, although I think that latter point may at least have been fixed in the meantime. But consistency with what Terraform itself does is the main thing here, so better to use the library it uses.

PR #44, opened 2023-08-08
Normalize relative paths to an absolute path

sebasslash:
## Description

This issue regarding dereferencing symlinks has plagued Terraform v1.3.0+ (reported for several versions [here](https://github.com/hashicorp/terraform/issues/31895)). The first issue, dereferencing symlinks whose targets resolve to symlinks _outside_ of the `src` directory was resolved in #37. The second issue, when the `src` directory itself is a symlink was resolved in #36. The last issue, and the focus of this PR, deals with normalizing paths when attempting to traverse the `src` directory. My change is simple: before attempting to traverse the `src` normalize the path representation to its absolute form. 

Big thank you to @willhughes-au and @calvinbui for providing the fixture used to test that paths are normalized. 

If you'd like to test that this works with Terraform, you will need to `go install .` from main, replacing the indirect dependency for go-slug with a repo containing these changes. 

```sh
cd /path/to/terraform
go mod edit -replace github.com/hashicorp/go-slug=/path/to/your/go-slug
go install .
```

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -158,6 +158,12 @@ func (p *Packer) Pack(src string, w io.Writer) (*Meta, error) {
 		ignoreRules = parseIgnoreFile(src)
 	}
 
+	// Ensure the source path provided is absolute
+	src, err = filepath.Abs(src)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read absolute path for source: %w", err)
+	}
+
 	// Walk the tree of files.
 	err = filepath.Walk(src, p.packWalkFn(src, src, src, tarW, meta, ignoreRules))
 	if err != nil {
@@ -264,7 +270,7 @@ func (p *Packer) packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta,
 			// If the target is a directory we can recurse into the target
 			// directory by calling the packWalkFn with updated arguments.
 			if resolved.info.IsDir() {
-				return filepath.Walk(resolved.absTarget, p.packWalkFn(root, resolved.target, path, tarW, meta, ignoreRules))
+				return filepath.Walk(resolved.absTarget, p.packWalkFn(root, resolved.absTarget, path, tarW, meta, ignoreRules))
 			}
 
 			// Dereference this symlink by updating the header with the target file
--- a/slug_test.go
+++ b/slug_test.go
@@ -84,6 +84,12 @@ func TestPack_absoluteSrcRelativeSymlinks(t *testing.T) {
 		// traversing the source directory
 		t.Fatalf("err: %v", err)
 	}
+
+	// Cannot pack without dereferencing
+	_, err = Pack(path, slug, false)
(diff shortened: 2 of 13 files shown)


slug.go
```diff
@@ -158,6 +158,13 @@ func (p *Packer) Pack(src string, w io.Writer) (*Meta, error) {
 		ignoreRules = parseIgnoreFile(src)
 	}
 
+	if !filepath.IsAbs(src) {
+		src, err = filepath.Abs(src)
+		if err != nil {
+			return nil, fmt.Errorf("something went super wrong root")
```
brandonc, line 164: ?

slug.go
```diff
@@ -183,6 +190,19 @@ func (p *Packer) packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta,
 			return err
 		}
 
+		if !filepath.IsAbs(src) {
+			src, err = filepath.Abs(src)
+			if err != nil {
+				return fmt.Errorf("failed to get absolute path for src %q: %w", src, err)
+			}
+		}
```
brandonc, line 198: Since src is a parameter of packWalkFn, does this need to be performed on every WalkFunc call?

brandonc (changes requested):

slug.go
sebasslash: I didn't update this error message

slug.go
```diff
@@ -158,6 +158,13 @@ func (p *Packer) Pack(src string, w io.Writer) (*Meta, error) {
 		ignoreRules = parseIgnoreFile(src)
 	}
 
+	if !filepath.IsAbs(src) {
```
brandonc, line 161: By my understanding, this check is redundant because filepath.Abs will return the path if it's already absolute, or else it will make it absolute using the current working directory.

slug_test.go
```diff
@@ -1088,8 +1122,8 @@ func assertArchiveFixture(t *testing.T, slug *bytes.Buffer, got *Meta) {
 		}
 
 		if hdr.Name == "example.tf" {
-			if hdr.Typeflag != tar.TypeSymlink {
-				t.Fatalf("expected symlink file 'example.tf'")
+			if hdr.Typeflag != tar.TypeReg {
+				t.Fatalf("expected symlink to be dereferenced 'example.tf'")
```
brandonc, line 1126: I feel like this change is representative of a change in behavior: When `dereference` is true, the godoc says "symlinks with a target outside of the src directory will be dereferenced." but it now appears that symlinks within the src directory are dereferenced as well.

brandonc (changes requested):

slug_test.go
sebasslash: In this case no, the previous test was incorrect. `example.tf` lies outside each `src` directory used in test, therefore it should be dereferenced.

brandonc: Wow!

brandonc (approved):

PR #47, opened 2023-10-11
Restore original timestamps when unpacking

brandonc:
## Description

The current implementation made no attempts to restore file, directory, or symlink info, including some permissions and all timestamps such that when the slug is hashed by something like [hashicorp/tfe's tfe_slug data source](https://github.com/hashicorp/terraform-provider-tfe/blob/main/internal/provider/data_source_slug.go#L44-L64), the hash of the source and unpacked destination were different.

- For files, the atime and mtime are restored after extraction
- For symlinks, only linux and darwin are supported because those are the only platforms that have the lutimes syscall. Other cross-platform go syscalls follow symlinks and it appears impossible to modify symlink timestamps on other platforms.
- For directories, the process of extracting files necessarily modifies the timestamps of the directory itself, so these are tracked in an internal data structure and then all the directories are modified after all files are extracted. This is similar to how tar preserves timestamps on directories. See https://www.gnu.org/software/tar/manual/html_node/Directory-Modification-Times-and-Permissions.html for details

I also refactored some of the unpacking rules and restoration methods into a new internal package because of the added complexity of platform dependency and build constraints. This helped to clean up some of the rules governing illegal slugs.

# Testing

I created a [special test repo](https://github.com/brandonc/example-go-slug-mtime) you can run against this package to test different example slugs.


internal/unpackinfo/unpackinfo_test.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			t.Errorf("failed to lstat %q: %s", info.Path, err)
+		}
+
+		if !info.IsSymlink() {
+			if stat.Mode() != info.OriginalMode {
+				t.Errorf("%q mode %q did not match expected header mode %q", info.Path, stat.Mode(), info.OriginalMode)
+			}
+		} else if CanMaintainSymlinkTimestamps() {
+			if !stat.ModTime().Truncate(time.Second).Equal(exampleModTime) {
```
ggambetti, line 142: I don't think we need to truncate here, I would expect the underlying system calls to set the exact time specified.

This is maybe impacted by the truncation in `lchtimes`?

slug.go
```diff
@@ -382,119 +392,92 @@ func (p *Packer) Unpack(r io.Reader, dst string) error {
...
-		if err := os.Chmod(path, mode); err != nil {
-			return fmt.Errorf("failed setting permissions on %q: %w", path, err)
+		if err := info.RestoreInfo(); err != nil {
+			return err
+		}
+	}
+
+	// Now that extraction is complete, restore mode and timestamps previously saved
+	// about directories.
+	for _, dir := range directoriesExtracted {
```
ggambetti, line 475: Are we guaranteed to have the directories occur in the right order here, or do we need to sort this list?

eg: `/foo/bar/` always before `/foo/`?

I know it would be possible to craft a file that would have a problem with this processing, and seeing the `os.MkdirAll` suggests that there aren't ordering guarantees.

internal/unpackinfo/lchtimes.go
```diff
@@ -0,0 +1,17 @@
...
+	"golang.org/x/sys/unix"
+)
+
+// Lchtimes modifies the access and modified timestamps on a target path
+// This capability is only available on Linux and Darwin as of now.
+func (i UnpackInfo) Lchtimes() error {
+	return unix.Lutimes(i.Path, []unix.Timeval{
+		{Sec: i.OriginalAccessTime.Unix()},
+		{Sec: i.OriginalModTime.Unix()}},
+	)
```
ggambetti, line 16: Both `unix.Timeval` and `time.Time` support more precision than Seconds, is there a particular reason we truncate here?

I've left some comments on the other files about time truncation that stem from this decision.

slug_test.go
```diff
@@ -1227,28 +1245,49 @@ func assertArchiveFixture(t *testing.T, slug *bytes.Buffer, got *Meta) {
...
+	}
+
+	dstInfo, err := os.Lstat(dst)
+	if err != nil {
+		t.Fatalf("dst file %q not found", dst)
+	}
+
+	// archive/tar purports to round timestamps to the nearest second so that behavior
+	// is duplicated here to test the restored timestamps.
+	sourceModTime := sourceInfo.ModTime().Round(time.Second)
```
ggambetti, line 1261: I think we want `Truncate` instead, I'd expect `Round` to be flakey?

```suggestion
	sourceModTime := sourceInfo.ModTime().Truncate(time.Second)
```

Ideally we wouldn't round/truncate here, because `lchtimes.go` would set the exact timestamp?

ggambetti (commented):
A few picky comments about the timestamp granularity we want, and wondering if directory iteration will work as expected.

I might also suggest that a defensive integration test be added (here or in `tfe_slug`) to catch future breaking changes that don't correctly round trip pack, unpack, repack.

internal/unpackinfo/unpackinfo.go
```diff
@@ -0,0 +1,165 @@
...
+		return i.restoreNormal()
+	}
+}
+
+func (i UnpackInfo) restoreDirectory() error {
+	if err := os.Chtimes(i.Path, i.OriginalAccessTime, i.OriginalModTime); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed setting times on directory %q: %w", i.Path, err)
+	}
+
+	if err := os.Chmod(i.Path, i.OriginalMode); err != nil && !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
brandonc: Wow! Good idea. I didn't see this show up in my testing but it seems like it should be ordered the other way around.

internal/unpackinfo/unpackinfo_test.go
brandonc: This is true! But only because there's no subseconds defined in the exampleModTime.

slug.go
brandonc: Good call out. I think this is only working based on how tree walk works during Pack.

slug_test.go
brandonc: Let's have all the time precision related discussion here!

By using Round, I was trying to [test this behavior](https://github.com/golang/go/blob/b86e8dd0f3a27bc9577690523c9feeb25593bec7/src/archive/tar/writer.go#L83-L92)

We _could_ modify our own code to truncate the second during Pack, but that would not be backwards-compatible with existing slugs.

Looking back at your other comment on Lchtimes, it seems like that, too, should use Round() instead of Unix(), to keep the same compatibility. I will give this another think tomorrow with a fresh brain.

slug.go
brandonc: ✅

internal/unpackinfo/unpackinfo_test.go
brandonc: ✅

internal/unpackinfo/unpackinfo.go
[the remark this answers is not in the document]
brandonc: ✅

internal/unpackinfo/unpackinfo.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// CanMaintainSymlinkTimestamps determines whether is is possible to change
+// timestamps on symlinks for the the current platform. For regular files
+// and directories, attempts are made to restore permissions and timestamps
+// after extraction. But for symbolic links, go's cross-platform
+// packages (Chmod and Chtimes) are not capable of changing symlink info
+// because those methods follow the symlinks. However, a platform-dependent option
+// is provided for linux and darwin (see Lchtimes)
+func CanMaintainSymlinkTimestamps() bool {
```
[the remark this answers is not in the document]
brandonc: ✅

slug_test.go
brandonc: OK with a fresh pair of eyes it seems to me that, if we presume that the slug was created by go-slug (not necessarily a safe assumption, but safe for the purposes of this PR) then Unix() is a safe way to express the times in Lchtimes, because the timestamp would have been rounded by the archive/tar header.

I've added some comments and review fixes in two new commits. Let me know if I've clarified the situation or if you think we'd better revisit that presumption!

brandonc: I suppose even if I chose to use subsecond precision during Lchtimes, it wouldn't matter since most of the time there would be no subseconds anyway!

ggambetti: Ah, so `archive/tar` complicates the situation. I'm inclined to suggest that we explicitly ignore the behaviour (except for how it complicates testing) and implement the changes as though we could always handle full accuracy or always truncate/round. That way `go-slug` is directly making the timestamp accuracy guarantee and doesn't depend on what `archive/tar` chooses to do.

slug.go
```diff
@@ -382,119 +398,98 @@ func (p *Packer) Unpack(r io.Reader, dst string) error {
...
 		}
 	}
+
+	// Now that extraction is complete, restore mode and timestamps previously saved
+	// about directories. But first, sort the list of directories by lexical order, which
+	// should impose a top-down ordering so that, in case subdirectories appear first in
+	// the archive, they won't change the modified timestamps of their parent directories.
+	sort.Slice(directoriesExtracted, func(i, j int) bool {
+		return directoriesExtracted[i].Path < directoriesExtracted[j].Path
+	})
```
ggambetti, line 485: I think this results in the wrong ordering, with `"foo"` before `"foo/bar"` which isn't what we want. Can we add a simple test for this?

internal/unpackinfo/lchtimes.go
ggambetti: Convo in the other thread about time.

ggambetti (commented):
A comment on handling directory time updates, but otherwise only more discussion on managing timestamp precision... 😅

ggambetti:
We had a conversation over Slack, which I'll try to summarize the result of it here:

I'm responsible for causing some confusion because I've gone back and forth on how we should handle timestamps: truncate/round or don't. After paying more attention to `archive/tar` and talking it through I think we can fix all our time accuracy problems by explicitly truncating/rounding up front. This should let us ignore the conditional `archive/tar` behaviour, simplify the `go-slug` timestamp preservation code, and not create a future bug where it looks like `go-slug` would preserve timestamp accuracy but isn't for some reason (because of `archive/tar`).

brandonc:
@Uk1288 Yes, that is expected if the files are part of the .terraformignore file

slug.go
brandonc: I don't think chtimes affects the parent directory mtime, because no matter how I order the tar headers for root and subdirectories, it doesn't seem to matter which order I restore the timestamps in. I created a test and a util to construct weird tar headers, but removed the sorting altogether.

ggambetti (commented):
LGTM.

PR #48, opened 2023-10-27
Add sourceaddrs.ParseFinalSource

alisdair:
We need to be able to decode string representations of final sources, and the existing ParseSource function doesn't handle versioned registry source addresses. This commit introduces ParseFileSource and supporting functions for decoding final registry sources.

### DIFF
--- a/sourceaddrs/source_final_test.go
+++ b/sourceaddrs/source_final_test.go
@@ -106,3 +106,86 @@ func TestResolveRelativeFinalSource(t *testing.T) {
 		})
 	}
 }
+
+func TestParseFinalSource(t *testing.T) {
+	onePointOh := versions.MustParseVersion("1.0.0")
+
+	tests := []struct {
+		Addr    string
+		Want    FinalSource
+		WantErr string
+	}{
+		{
+			Addr: "./a/b",
+			Want: MustParseSource("./a/b").(FinalSource),
+		},
+		{
+			Addr: "git::https://github.com/hashicorp/go-slug.git//beep/boop",
+			Want: MustParseSource("git::https://github.com/hashicorp/go-slug.git//beep/boop").(FinalSource),
+		},
+		{
+			Addr: "git::https://github.com/hashicorp/go-slug.git//beep@1.2.3/boop",
+			Want: MustParseSource("git::https://github.com/hashicorp/go-slug.git//beep@1.2.3/boop").(FinalSource),
+		},
+		{
+			Addr: "example.com/foo/bar/baz@1.0.0//beep",
+			Want: MustParseSource("example.com/foo/bar/baz//beep").(RegistrySource).Versioned(onePointOh),
+		},
+		{
+			Addr: "example.com/foo/bar/baz@1.0.0",
+			Want: MustParseSource("example.com/foo/bar/baz").(RegistrySource).Versioned(onePointOh),
+		},
+		{
+			Addr: "gitlab.com/hashicorp/go-slug/bleep@1.0.0",
+			Want: MustParseSource("gitlab.com/hashicorp/go-slug/bleep").(RegistrySource).Versioned(onePointOh),
+		},
+		{
+			Addr: "./a/b@1.0.0",
+			Want: MustParseSource("./a/b@1.0.0").(FinalSource),
+		},
+		{
+			Addr:    " ./a/b",
+			WantErr: "source address must not have leading or trailing spaces",
+		},
+		{
+			Addr:    "",
+			WantErr: "a valid source address is required",
+		},
+		{
+			Addr:    "example.com/foo/bar/baz@1.0.x//beep",
+			WantErr: `invalid module registry source address "example.com/foo/bar/baz@1.0.x//beep": invalid version: can't use wildcard for patch number; an exact version is required`,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.Addr, func(t *testing.T) {
+			got, gotErr := ParseFinalSource(test.Addr)
+
+			if test.WantErr != "" {
+				if gotErr == nil {
+					t.Fatalf("unexpected success\ngot result: %#v (%T)\nwant error: %s", got, got, test.WantErr)
+				}
+				if got, want := gotErr.Error(), test.WantErr; got != want {
+					t.Fatalf("wrong error\ngot error:  %s\nwant error: %s", got, want)
+				}
+				return
+			}
+
+			if gotErr != nil {
+				t.Fatalf("unexpected error: %s", gotErr)
+			}
+
+			// Two addresses are equal if they have the same string representation
+			// and the same dynamic type.
+			gotStr := got.String()
+			wantStr := test.Want.String()
+			if gotStr != wantStr {
+				t.Errorf("wrong result\ngot:  %s\nwant: %s", gotStr, wantStr)
+			}
+
+			if gotType, wantType := reflect.TypeOf(got), reflect.TypeOf(test.Want); gotType != wantType {
+				t.Errorf("wrong result type\ngot:  %s\nwant: %s", gotType, wantType)
+			}
+		})
+	}
+}
--- a/sourceaddrs/source_registry_final.go
+++ b/sourceaddrs/source_registry_final.go
@@ -1,6 +1,9 @@
 package sourceaddrs
 
 import (
+	"fmt"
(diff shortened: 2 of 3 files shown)


sourceaddrs/source_registry_final.go
```diff
@@ -24,9 +27,39 @@ type RegistrySourceFinal struct {
...
+		pkg = matches[1]
+		path = matches[3]
+	}
+	return looksLikeRegistrySource(fmt.Sprintf("%s//%s", pkg, path))
+}
+
 // finalSourceSigil implements FinalSource
 func (s RegistrySourceFinal) finalSourceSigil() {}
 
+// ParseRemoteSource parses the given string as a final registry source
```
apparentlymart, line 42: The name in the comment doesn't match the name of the function. 😀

sourceaddrs/source_registry_final.go
```diff
@@ -62,3 +95,8 @@ func (s RegistrySourceFinal) FinalSourceAddr(realSource RemoteSource) RemoteSour
 	// paths together, so we can just delegate to our unversioned equivalent.
 	return s.Unversioned().FinalSourceAddr(realSource)
 }
+
+// finalRegistrySourcePattern is a non-exhaustive regexp which looks only for
+// the expected three components of a RegistrySourceFinal string encoding: the
+// package address, version, and subpath.
+var finalRegistrySourcePattern = regexp.MustCompile(`^(.+)@(.+)//(.+)$`)
```
apparentlymart, line 102: Although not true for all of the single-file addresses you've been dealing with in your current work, it's possible to have a "final address" without a sub-path portion if it refers to the root directory of the registry module package.

Therefore I think the part of this pattern starting with `//` ought to be optional, although I suspect doing that will make the pattern highly ambiguous (it will essentially just be "is there at least one at sign?") so we might also need to extend the "looks like final registry source" function to actually try parsing the version part as a version number, before concluding that it's safe to discard it and see if what's left is a valid non-final registry source. 🤔

sourceaddrs/source_final_test.go
```diff
@@ -106,3 +106,70 @@ func TestResolveRelativeFinalSource(t *testing.T) {
...
+		},
+		{
+			Addr:    "",
+			WantErr: "a valid source address is required",
+		},
+		{
+			Addr:    "example.com/foo/bar/baz@1.0.x//beep",
+			WantErr: `invalid module registry source address "example.com/foo/bar/baz@1.0.x//beep": invalid version: can't use wildcard for patch number; an exact version is required`,
+		},
+	}
```
apparentlymart, line 142: I suggest also some local and remote addresses that have at signs in their path portions, to make sure they don't get misclassified as being registry addresses.

For example, `./foo/bar/baz@1.0.0` should presumably match as a local source whose final path segment happens to contain an at sign. I expect that test will already pass but I'd like to ensure it doesn't get broken in future maintenance, since this sort of thing was already a maintenance hazard for the similar code in Terraform Core's `addrs` package.

apparentlymart (commented):
This approach of reusing as much as possible from the non-final parsing code seems plausible to me but I think the regex pattern is not quite right to accept all valid input.

sourceaddrs/source_registry_final.go
alisdair: Whoops! Fixed, thanks.

sourceaddrs/source_registry_final.go
alisdair: Ah, thanks for spotting this oversight. I've updated the pattern and the code using it, with one caveat:

> we might also need to extend the "looks like final registry source" function to actually try parsing the version part as a version number

I'm not sure about this part. Is this from a "what if?" cautious stance, or are there ambiguities in the final source address format I'm not seeing that make it necessary?

I intend the regexp only to help split the three (or two!) components of the final registry source, and lean on the registry address parser (ignoring the version) to check if the string looks like a final registry source. I don't think it's possible for any other source address type to match the regexp, and then have its package and subpath parse as a registry source address.

The reason I'm asking is because making an address like `example.com/foo/bar/baz@1.0.x` fail to match `looksLikeFinalRegistrySource` means that it will instead be interpreted as a remote source. As a result we won't see a useful error explaining that the version is invalid.

sourceaddrs/source_final_test.go
alisdair: Done!

sourceaddrs/source_registry_final.go
apparentlymart: I must confess that I am coming at this from a cautious standpoint because in the similar code over in Terraform Core we initially had a not-quite-tight-enough definition of "is a registry source address" which caused Terraform to misinterpret some addresses that were intended to be remote source addresses, causing confusing error messages. That was hard to resolve retroactively because we needed to avoid breaking the handling of any addresses which the author _did_ intend to be a registry source address. :grimacing:  (There's some commentary resulting from that earlier mess [in the Terraform Core `addrs` package](https://github.com/hashicorp/terraform/blob/4b1c3457a4090235645b699ec360ff9a7a2f4838/internal/addrs/module_source.go#L83-L96).

With that said, I would be happy if we can convince ourselves that the concern doesn't apply here by adding additional test cases, rather than adding additional code.

---

You have pointed out something in your response which I think is the same potential problem viewed from the opposite angle: the goal of this function is to try to differentiate between user intent to write a registry source and user intent to write a remote source. If we misinterpret the user's intent in either direction then the result will be a confusing error message. I'm worrying about misclassifying a remote source as a registry source, and you're worrying about the converse, but a misclassification in either direction is equally confusing.

On further reflection, I think we might be saved here by this implementation being only a subset of what `go-getter` would accept, meaning that we don't have quite as many highly-ambiguous forms as Terraform itself currently needs to deal with. I used the tests for `ParseSource` to remind myself what's included in the subset that `go-slug` implements and I think the only ones that seem worthy of some careful scrutiny are these shorthand `github.com` and `gitlab.com` variants, which are syntactically close to the registry address syntax:

https://github.com/hashicorp/go-slug/blob/aeb5117bdbb942fec6a3308de57ecf96191316c9/sourceaddrs/source_test.go#L213-L273

So some examples of what might be misclassified by an unsufficiently-specific rule could include:

* `github.com/foo/bar/baz@1.0.0` (a directory called `baz@1.0.0` inside a GitHub repository)
* `github.com/foo/bar//baz@1.0.0` (the previous but written using the explicit sub-path syntax)

The last case in the range I linked above is also pretty :grimacing: -- `gitlab.com` is _both_ a Git repository host and a Terraform module registry, and so the handling of that is _unavoidably_ ambiguous. If we are parsing something that could be either then it seems we prefer the module registry interpretation and require `git::` prefix to access the other interpretation. So I suppose we should make sure we still resolve that ambiguity in the same way after the registry interpretation has becomes a "final" registry address.

alisdair: I think our disconnect here stems from this:

> the goal of this function is to try to differentiate between user intent to write a registry source and user intent to write a remote source

My intent with this function wasn't intended to handle user input. It's intended to parse the encoded output of previously-parsed user input. But since there's nothing _stopping_ a user of this library from running it on user input, I can see the concern now.

As of my last commit, the first example above has a direct equivalent in the tests, which I think works as you'd expect: https://github.com/hashicorp/go-slug/blob/7ff3155b01cf46c00449e272b8dc90b1e80c8c06/sourceaddrs/source_final_test.go#L134-L137

The second example above currently doesn't parse at all, either with `ParseSource` or `ParseFinalSource`:

> `source_test.go:428: unexpected error: invalid remote source address "gitlab.com/hashicorp/go-slug//bleep/bloop": invalid sub-path: must be slash-separated relative path without any .. or . segments`

I confess I'm rather turned around at this point, and I'm not sure what changes you're recommending. If you could let me know what your current thinking is I'll be happy to implement whatever you suggest.

apparentlymart: Those examples failing to parse at all is also a fine outcome. I was concerned only that they might be misunderstood as being of the wrong type.

I'm going to now re-read the tests as you currently have them with fresh eyes and brain.

apparentlymart: I think the current suite of tests you have is sufficient to show that any ambiguities in the regex are only hypothetical problems, and not actual problems. :+1:

apparentlymart (approved):

PR #51, opened 2023-11-28
Skip processing deleted directories during sourcebundle.Build

liamcervante:
Previously, when building the source bundle, if an excluded directory had files in it, the build process would error because filepath.Walk computes the list of children for a directory before executing the walk function on that directory. Our function was deleting the directory (and all children) and then returning an error later when it tried to process (the now deleted) children.

Now, when we delete the directory we return the filepath.SkipDir directory which tells it to stop trying to recurse and it won't process the deleted children.

### DIFF
--- a/sourcebundle/builder.go
+++ b/sourcebundle/builder.go
@@ -13,10 +13,11 @@ import (
 	"sync"
 
 	"github.com/apparentlymart/go-versions/versions"
-	"github.com/hashicorp/go-slug/internal/ignorefiles"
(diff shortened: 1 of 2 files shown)


apparentlymart (approved):
This seems good to me!

There's some parallel discussion in https://github.com/hashicorp/go-slug/pull/50 about how this strategy is also not handling exclusions that descend from already-ignored directories, which I think is separate from what you fixed here but I just want to connect the two with links in case we want to refer back to these discussions later.

PR #53, opened 2024-01-18
Add `NormalizeUnpackFileModes` option to `Packer` struct

sebasslash:
A recent change was introduced that preserves a file's original atime, mtime and mode when unpacked from a tarball. This consequently shifted responsibility to the slug's consumer to ensure each untarred file has the expected permissions. Inevitably this can lead to unpredictable errors, particularly if slugs are created by unknown sources and are expected to be handled uniformly. 

This PR aims to add a new packer option, `NormalizeUnpackFileModes`, that configures go-slug to normalize file modes when unpacking configuration files. This change aims to behave closely to how git normalizes permissions. The permissions for each type of file are encoded as follows:
- Regular, non-executable: 0644
- Regular, executable: 0755
- Directory: 040755
- Symlink: 0120777

All other file types are unsupported and result in an error: named pipes, char devices, sockets, etc. 

Example Usage:

```go
p, err := NewPacker(NormalizeUnpackFileModes(), DereferenceSymlinks())
if err != nil { // do a thing }

// Omitting code to handle packing or loading slug
if err = p.Unpack("/path/to/slug", "/path/to/destination"); err != nil {
  panic("whoops we made a booboo")
}
```

### DIFF
--- a/internal/unpackinfo/filemode.go
+++ b/internal/unpackinfo/filemode.go
@@ -53,7 +53,7 @@ func (m FileMode) ToFsFileMode() (fs.FileMode, error) {
 	case Plain:
 		return fs.FileMode(0644), nil
 	case Dir:
-		return fs.ModePerm | fs.ModeDir, nil
+		return fs.FileMode(0755) | fs.ModeDir, nil
 	case Executable:
 		return fs.FileMode(0755), nil
 	case Symlink:
--- a/internal/unpackinfo/filemode.go
+++ b/internal/unpackinfo/filemode.go
@@ -12,9 +12,9 @@ const (
 	Empty FileMode = 0
 	// Dir represent a Directory.
 	Dir FileMode = 0040000
-	// Regular represent non-executable files.  Please note this is not
+	// Plain represent non-executable, text files.  Please note this is not
 	// the same as golang regular files, which include executable files.
-	Regular FileMode = 0100644
+	Plain FileMode = 0100644
 	// Executable represents executable files.
 	Executable FileMode = 0100755
 	// Symlink represents symbolic links to files.
@@ -32,7 +32,7 @@ func NewFileMode(mode fs.FileMode) (FileMode, error) {
 			return Executable, nil
 		}
 
-		return Regular, nil
+		return Plain, nil
 	}
 
 	if mode.IsDir() {
@@ -50,7 +50,7 @@ func NewFileMode(mode fs.FileMode) (FileMode, error) {
 // permissions for regular and executable files.
 func (m FileMode) ToFsFileMode() (fs.FileMode, error) {
 	switch m {
-	case Regular:
+	case Plain:
 		return fs.FileMode(0644), nil
 	case Dir:
 		return fs.ModePerm | fs.ModeDir, nil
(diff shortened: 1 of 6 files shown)


internal/unpackinfo/filemode.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+type FileMode uint32
+
+const (
+	Empty FileMode = 0
+	// Dir represent a Directory.
+	Dir FileMode = 0040000
+	// Regular represent non-executable files.  Please note this is not
+	// the same as golang regular files, which include executable files.
```
nfagerlund, line 16: Hmm. Do you think it's worth using a slightly different name than Go does to avoid confusion? If so, I suggest "Plain".

internal/unpackinfo/filemode.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// Maps a FileMode integer to an fs.FileMode type, normalizing
+// permissions for regular and executable files.
+func (m FileMode) ToFsFileMode() (fs.FileMode, error) {
+	switch m {
+	case Regular:
+		return fs.FileMode(0644), nil
+	case Dir:
+		return fs.ModePerm | fs.ModeDir, nil
```
nfagerlund, line 56: This made me doubt myself and I ran in circles about it for a while, but! I think we actually want directories to be 755 (rwx r-x r-x), not 777 (world-writeable). Thoughts?

internal/unpackinfo/unpackinfo.go
```diff
@@ -68,49 +70,59 @@ func NewUnpackInfo(dst string, header *tar.Header) (UnpackInfo, error) {
...
 
 // IsSymlink describes whether the file being unpacked is a symlink
-func (i UnpackInfo) IsSymlink() bool {
+func (i *UnpackInfo) IsSymlink() bool {
 	return i.Typeflag == tar.TypeSymlink
 }
 
 // IsDirectory describes whether the file being unpacked is a directory
-func (i UnpackInfo) IsDirectory() bool {
+func (i *UnpackInfo) IsDirectory() bool {
```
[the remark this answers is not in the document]
sebasslash: Normally I've followed the styling that disallows mixing receivers ([see Go's guidance](https://go.dev/tour/methods/8#:~:text=In%20general%2C%20all%20methods%20on,not%20a%20mixture%20of%20both.)). So given I needed a pointer receiver to configure an instance of `UnpackInfo`, I changed all of the methods for consistency.

internal/unpackinfo/filemode.go
sebasslash: In truth I followed the permissions set by [go-git](https://github.com/go-git/go-git/blob/9a6715aba7a88defd27aeba9ddb24cce9a9416c1/plumbing/filemode/filemode.go#L173). I wonder if there is a specific use case in git that requires dirs with r+w+x for all groups/users. I think for the purposes of Terraform configuration this would be irrelevant so we can get by with 0755. I will double check this to be sure.

internal/unpackinfo/unpackinfo.go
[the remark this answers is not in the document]
nfagerlund: ah, okay! Solid answer, thank you.

internal/unpackinfo/filemode.go
nfagerlund: 🤔 I wonder if there's a mask being applied somewhere later down the chain. Maybe do a real-world test and see what comes out; this is a case where I'm dubious about trusting what the code claims it's doing.

internal/unpackinfo/unpackinfo.go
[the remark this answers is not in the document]
brandonc: TIL

brandonc:
I don't feel this needs to be a new option. Why wouldn't we make it the default behavior?

PR #54, opened 2024-01-30
Static defaultExclusions rules are not congruent

brandonc:
The default rule set did not reflect the necessary "negationsAfter" state for rules that appear before the default negation rule.

This caused Pack to skip the entire ".terraform/modules" directory when it was encountered because it believed the rule to be dominant.

Any time a .terraformignore file containing a negation was evaluated, these default rules would be corrected by the parsing algorithm. This is why this bug was not caught by the tests. So I added another Pack test that relies only on the default rules.

In the second commit, moving the .git/ exclusion rule to the end of the default list, makes it possible to skip the entire .git/ directory walk because no exclusions appear after it. This should slightly speed up packing performance when the default rules are used or when no additional exclusion rules are set by .terraformignore.

The field `dirs` on each rule was not used so I removed it entirely.


alisdair (approved):
Nice! Change looks good, and I verified that a build of Terraform using this fixed the issue for me.

PR #56, opened 2024-04-15
Collect VCS metadata

matejrisek:
Collect git metadata - commit ID and commit message and store it to metadata

This PR is mostly to garner feedback, however if the approach is good enough it'd be great to merge it.

Related to https://github.com/hashicorp/tfc-agent/pull/666/files

### DIFF
--- a/sourcebundle/builder.go
+++ b/sourcebundle/builder.go
@@ -561,11 +561,11 @@ func (b *Builder) writeManifest(filename string) error {
 			LocalDir:   localDirName,
 		}
 		if pkgMeta != nil {
-			if pkgMeta.GitCommitID != "" {
-				manifestPkg.Meta.GitCommitID = pkgMeta.GitCommitID
+			if pkgMeta.gitCommitID != "" {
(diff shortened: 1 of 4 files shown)


apparentlymart (approved):
This seems reasonable to me!

Although this isn't urgent to figure out right now while this package is still early anyway, it might be worth at some point documenting exactly what form of commit message the fetchers are expected to return -- e.g. whether it's the full message verbatim, or just the summary line -- so that it's clearer which part of the system is responsible for reducing to just the summary once that inevitably arises in some parts of the UI.

matejrisek:
> This seems reasonable to me!
> 
> Although this isn't urgent to figure out right now while this package is still early anyway, it might be worth at some point documenting exactly what form of commit message the fetchers are expected to return -- e.g. whether it's the full message verbatim, or just the summary line -- so that it's clearer which part of the system is responsible for reducing to just the summary once that inevitably arises in some parts of the UI.

Makes sense, will note that down. We're planning on doing the refactoring once we're done with the base functionality work.

PR #59, opened 2024-05-14
Expose deprecation data in Bundle

Maed223:
Updates sourceBundle’s `ModulePackageVersionsResponse` type to include registry module deprecation data in the response. This deprecation data is then propagated to the Bundle, where it will be then used within a related `tfc-agent` PR to add diagnostic warnings about deprecated module versions for stacks .

### Related Items
- [Jira Ticket](https://hashicorp.atlassian.net/browse/TF-14020?atlOrigin=eyJpIjoiYWQyZDAzYjkyYzdkNDdkZGEyZDk0YTFlODExZWVmYWMiLCJwIjoiaiJ9)
- [`tfc-agent` PR](https://github.com/hashicorp/tfc-agent/pull/687)

### DIFF
--- a/sourceaddrs/source_registry.go
+++ b/sourceaddrs/source_registry.go
@@ -29,6 +29,12 @@ type RegistrySource struct {
 	subPath string
 }
 
+type RegistryVersionDeprecation struct {
+	Version string
+	Reason  string
+	Link    string
+}
+
 // sourceSigil implements Source
 func (s RegistrySource) sourceSigil() {}
 
--- a/sourceaddrs/source_registry.go
+++ b/sourceaddrs/source_registry.go
@@ -29,12 +29,6 @@ type RegistrySource struct {
 	subPath string
 }
 
-type RegistryVersionDeprecation struct {
-	Version string
-	Reason  string
-	Link    string
-}
-
 // sourceSigil implements Source
 func (s RegistrySource) sourceSigil() {}
 
--- a/sourceaddrs/source_remote.go
+++ b/sourceaddrs/source_remote.go
@@ -15,6 +15,17 @@ type RemoteSource struct {
 	subPath string
 }
 
+type RemoteSourceInfo struct {
+	RemoteSource       RemoteSource
+	VersionDeprecation *Deprecation
+}
+
+type Deprecation struct {
+	Version string
+	Reason  string
+	Link    string
+}
+
 var _ Source = RemoteSource{}
 var _ FinalSource = RemoteSource{}
 
--- a/sourceaddrs/source_remote.go
+++ b/sourceaddrs/source_remote.go
@@ -15,17 +15,6 @@ type RemoteSource struct {
 	subPath string
 }
 
-type RemoteSourceInfo struct {
-	RemoteSource       RemoteSource
-	VersionDeprecation *Deprecation
-}
-
-type Deprecation struct {
-	Version string
-	Reason  string
-	Link    string
-}
-
 var _ Source = RemoteSource{}
 var _ FinalSource = RemoteSource{}
 
--- a/sourcebundle/builder.go
+++ b/sourcebundle/builder.go
@@ -71,7 +71,9 @@ type Builder struct {
 
 	// resolvedRegistry tracks the underlying remote source address for each
 	// selected version of each module registry package.
-	resolvedRegistry map[registryPackageVersion]sourceaddrs.RemoteSourceInfo
+	resolvedRegistry map[registryPackageVersion]sourceaddrs.RemoteSource
+
+	resolvedRegistryVersionDeprecations map[registryPackageVersion]*sourceaddrs.RegistryVersionDeprecation
 
 	// registryPackageVersions caches responses from module registry calls to
 	// look up the available versions for a particular module package. Although
@@ -100,14 +102,15 @@ func NewBuilder(targetDir string, fetcher PackageFetcher, registryClient Registr
 		return nil, fmt.Errorf("invalid target directory: %w", err)
 	}
 	return &Builder{
-		targetDir:               absDir,
-		fetcher:                 fetcher,
-		registryClient:          registryClient,
-		analyzed:                make(map[remoteArtifact]struct{}),
-		remotePackageDirs:       make(map[sourceaddrs.RemotePackage]string),
-		remotePackageMeta:       make(map[sourceaddrs.RemotePackage]*PackageMeta),
-		resolvedRegistry:        make(map[registryPackageVersion]sourceaddrs.RemoteSourceInfo),
-		registryPackageVersions: make(map[regaddr.ModulePackage][]ModulePackageInfo),
+		targetDir:                           absDir,
+		fetcher:                             fetcher,
+		registryClient:                      registryClient,
+		analyzed:                            make(map[remoteArtifact]struct{}),
+		remotePackageDirs:                   make(map[sourceaddrs.RemotePackage]string),
+		remotePackageMeta:                   make(map[sourceaddrs.RemotePackage]*PackageMeta),
+		resolvedRegistry:                    make(map[registryPackageVersion]sourceaddrs.RemoteSource),
+		resolvedRegistryVersionDeprecations: make(map[registryPackageVersion]*sourceaddrs.RegistryVersionDeprecation),
+		registryPackageVersions:             make(map[regaddr.ModulePackage][]ModulePackageInfo),
 	}, nil
 }
 
@@ -384,7 +387,7 @@ func (b *Builder) findRegistryPackageSource(ctx context.Context, sourceAddr sour
 		pkg:     pkgAddr,
 		version: selectedVersion,
 	}
-	realSourceInfo, ok := b.resolvedRegistry[pkgVer]
+	realSourceAddr, ok := b.resolvedRegistry[pkgVer]
 	if !ok {
 		var reqCtx context.Context
 		if cb := trace.RegistryPackageSourceStart; cb != nil {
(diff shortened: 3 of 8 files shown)


sourceaddrs/source_remote.go
```diff
@@ -15,6 +15,17 @@ type RemoteSource struct {
 	subPath string
 }
 
+type RemoteSourceInfo struct {
+	RemoteSource       RemoteSource
+	VersionDeprecation *Deprecation
```
brandonc, line 20: I think I'm unsatisfied that these new structs appear in this package but are not used by this package at all.

We should either change ParseRemoteSource/MakeRemoteSource to include deprecation data in order to simplify sourcebundle data structures... or we should move these types into sourcebundle and wrap RemoteSource as normal. I think I am in favor of the latter since deprecation data isn't a "source address"

brandonc (commented):
Got a couple concerns with this, but they are minor. We should find a way to create a meaningful test

sourcebundle/builder.go
```diff
@@ -713,3 +733,20 @@ func packagePrepareWalkFn(root string, ignoreRules *ignorefiles.Ruleset) filepat
 		return nil
 	}
 }
+
+func extractVersionListFromResponse(modPackageInfos []ModulePackageInfo) versions.List {
+	vs := make(versions.List, len(modPackageInfos))
+	for _, v := range modPackageInfos {
+		vs = append(vs, v.Version)
+	}
```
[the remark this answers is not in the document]
Maed223: Ah that makes perfect sense, thanks for the go tip! 

From what I can surmise it seems that either approach doesn't have a specific benefit over the other in this instance. Since we know that in the end the length of the slice will always be equal to its capacity. So allocating memory for all the element up front, as in the first suggestion, vs doing so within the loop with append in total doesn't make a difference. Is my thinking correct?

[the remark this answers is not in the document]
brandonc: The only difference it makes is that allocating the space up front saves time and memory because it doesn't require that the slice grow after some appends happen. When a slice grows, it can allocate not enough or too much as well as copy all the elements from the old slice to the new, larger one. If you know the size slice you need, it's usually better to use the first option.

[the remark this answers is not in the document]
Maed223: Gotcha, thanks for the more detailed explanation. Included 👍 https://github.com/hashicorp/go-slug/pull/59/commits/3fa65750f5638397e8c92aa9bc469e4d7d628c52

sourceaddrs/source_remote.go
Maed223: Ah when you put it like that I have to totally agree. Included 👍 https://github.com/hashicorp/go-slug/pull/59/commits/3fa65750f5638397e8c92aa9bc469e4d7d628c52

sourceaddrs/source_registry.go
```diff
@@ -29,6 +29,12 @@ type RegistrySource struct {
 	subPath string
 }
 
+type RegistryVersionDeprecation struct {
```
brandonc, line 32: This type may be in the wrong package, too.

brandonc: I like the way you un-nested the thing previously called "RemoteSourceInfo"!

Maed223: Ah gotcha. Just to clarify your thinking, would you consider it out of place since it's only used in `sourcebundle`?

Maed223: And in any case any suggestions where it might be a better fit? No file in particular pops out at me in the package.

Maed223: Ended up moving it to `package_meta.go` in the sourcebundle package, seems to be the best fit.

sourcebundle/builder.go
```diff
@@ -100,14 +102,15 @@ func NewBuilder(targetDir string, fetcher PackageFetcher, registryClient Registr
...
-		resolvedRegistry:        make(map[registryPackageVersion]sourceaddrs.RemoteSource),
-		registryPackageVersions: make(map[regaddr.ModulePackage]versions.List),
+		targetDir:                           absDir,
+		fetcher:                             fetcher,
+		registryClient:                      registryClient,
+		analyzed:                            make(map[remoteArtifact]struct{}),
+		remotePackageDirs:                   make(map[sourceaddrs.RemotePackage]string),
+		remotePackageMeta:                   make(map[sourceaddrs.RemotePackage]*PackageMeta),
+		resolvedRegistry:                    make(map[registryPackageVersion]sourceaddrs.RemoteSource),
+		resolvedRegistryVersionDeprecations: make(map[registryPackageVersion]*RegistryVersionDeprecation),
```
brandonc, line 112: Slightly verbose-- plus 'resolved' may refer to the RemoteSource values...

```suggestion
		versionDeprecations: make(map[registryPackageVersion]*RegistryVersionDeprecation),
```

sourcebundle/bundle.go
```diff
@@ -354,6 +362,11 @@ func (b *Bundle) RegistryPackageVersions(pkgAddr regaddr.ModulePackage) versions
 	return ret
 }
 
+func (b *Bundle) RegistryPackageVersionDeprecation(pkgAddr regaddr.ModulePackage, version versions.Version) *RegistryVersionDeprecation {
+	deprecation := b.registryPackageVersionDeprecations[pkgAddr][version]
+	return deprecation
```
brandonc, line 367: I thought this could panic, but it wont! Neat. No need for a local, though

```suggestion
	return b.registryPackageVersionDeprecations[pkgAddr][version]
```

sourcebundle/builder_test.go
```diff
@@ -593,6 +686,23 @@ func testingBuilder(t *testing.T, targetDir string, remotePackages map[string]st
...
 	}
 
+	for pkgAddrRaw, deprecations := range registryVersionDeprecations {
+		pkgAddr, err := sourceaddrs.ParseRegistryPackage(pkgAddrRaw)
+		if err != nil {
+			t.Fatalf("invalid registry package address %q: %s", pkgAddrRaw, err)
+		}
+		if registryDeprecations[pkgAddr.Namespace] == nil {
+			registryDeprecations[pkgAddr.Namespace] = make(map[versions.Version]*ModulePackageVersionDeprecation)
+		}
```
brandonc, line 696: I don't think you need to reallocate this nested map: https://go.dev/play/p/dmUQYnvF7y_H

sourcebundle/bundle.go
```diff
@@ -108,11 +110,17 @@ func OpenDir(baseDir string) (*Bundle, error) {
 			vs = make(map[versions.Version]sourceaddrs.RemoteSource)
 			ret.registryPackageSources[pkgAddr] = vs
 		}
+		deprecations := ret.registryPackageVersionDeprecations[pkgAddr]
+		if deprecations == nil {
+			deprecations = make(map[versions.Version]*RegistryVersionDeprecation)
```
brandonc, line 115: Unnecessary reallocation I think (see previous comment)

brandonc (commented):
Really basic feedback. I didn't smoke test but looks safe

sourcebundle/builder_test.go
Maed223: That's what I thought initially as well, but I'm otherwise getting nil pointer exceptions. Any idea why that could be the case?

sourcebundle/bundle.go
Maed223: Similar to the above, I'm getting nil pointer exceptions without it.

sourcebundle/builder_test.go
Maed223: Actually nevermind that 😅, didn't initially catch that you were assigning a map at the index.

sourcebundle/bundle.go
Maed223: I suppose that similar to the above we could just assign the nested map at the index of `pkgAddr` but I figured to was best to stick with the style of how `registryPackageSources` is handled above for consistencies sake.

```go
vs := ret.registryPackageSources[pkgAddr]
if vs == nil {
    vs = make(map[versions.Version]sourceaddrs.RemoteSource)
    ret.registryPackageSources[pkgAddr] = vs
}
```

sourcebundle/builder_test.go
Maed223: Included 👍  https://github.com/hashicorp/go-slug/pull/59/commits/ea8f696aaf4615b5eed02231bed0038ff283c397

sourcebundle/bundle.go
Maed223: Included 👍  https://github.com/hashicorp/go-slug/pull/59/commits/ea8f696aaf4615b5eed02231bed0038ff283c397

sourcebundle/builder.go
Maed223: Could we meet in the middle with `packageVersionDeprecations`? 

Included here: https://github.com/hashicorp/go-slug/pull/59/commits/ea8f696aaf4615b5eed02231bed0038ff283c397

sourcebundle/builder.go
```diff
@@ -73,12 +73,14 @@ type Builder struct {
 	// selected version of each module registry package.
 	resolvedRegistry map[registryPackageVersion]sourceaddrs.RemoteSource
```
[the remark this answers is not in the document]
Maed223: Included 👍  https://github.com/hashicorp/go-slug/pull/59/commits/ea8f696aaf4615b5eed02231bed0038ff283c397

sourcebundle/builder.go
```diff
@@ -73,12 +73,16 @@ type Builder struct {
...
 
+	// resolvedRegistryVersionDeprecations tracks potential deprecations for
+	// each package version
+	packageVersionDeprecations map[registryPackageVersion]*RegistryVersionDeprecation
```
alisdair, line 78: I'm curious why this is a map of pointer-to-`RegistryVersionDeprecation`. Later on we store `nil` in this map if a registry package version has no deprecation information, rather than having no entry in the map at all. Is that important?

Having a pointer here means we need to be diligent about nil checks, or we'll panic. Would there be a disadvantage to this being `map[registryPackageVersion]RegistryVersionDeprecation` instead?

sourcebundle/builder.go
```diff
@@ -73,12 +73,16 @@ type Builder struct {
 	// selected version of each module registry package.
 	resolvedRegistry map[registryPackageVersion]sourceaddrs.RemoteSource
 
+	// resolvedRegistryVersionDeprecations tracks potential deprecations for
+	// each package version
```
alisdair, line 77: This comment is currently just restating the variable name, and some more detail would be helpful. Example future readers' questions you may want to proactively answer (or not! whatever you think is pertinent):

- Where is the data gathered from?
- What consumes this data?
- What's a "potential" deprecation?

While we're here, the field in the comment doesn't match the field name in the struct.

sourcebundle/builder.go
```diff
@@ -73,12 +73,16 @@ type Builder struct {
...
+	packageVersionDeprecations map[registryPackageVersion]*RegistryVersionDeprecation
+
 	// registryPackageVersions caches responses from module registry calls to
 	// look up the available versions for a particular module package. Although
 	// these could potentially change while we're running, we assume that the
 	// lifetime of a particular Builder is short enough for that not to
 	// matter.
```
alisdair, line 84: Might be worth noting here that this cache includes the deprecation data, as well as the package versions.

sourcebundle/builder.go
```diff
@@ -713,3 +736,20 @@ func packagePrepareWalkFn(root string, ignoreRules *ignorefiles.Ruleset) filepat
...
+func extractVersionListFromResponse(modPackageInfos []ModulePackageInfo) versions.List {
+	vs := make(versions.List, len(modPackageInfos))
+	for index, v := range modPackageInfos {
+		vs[index] = v.Version
+	}
+	vs.Sort()
+	return vs
+}
+
+func extractVersionDeprecationsFromResponse(modPackageInfos []ModulePackageInfo) map[versions.Version]*ModulePackageVersionDeprecation {
```
alisdair, line 749: This function is very simple and only called from one site. I think it's more typical go style to inline this instead.

alisdair (approved):
Just some minor go-specific thoughts. From a stacks technical point of view, this looks fine to me!

sourcebundle/builder.go
Maed223: Included 👍 https://github.com/hashicorp/go-slug/pull/59/commits/be5d90781af6a49d0b5f5dc5fca24e185629c14a

sourcebundle/builder.go
Maed223: Included 👍 https://github.com/hashicorp/go-slug/pull/59/commits/be5d90781af6a49d0b5f5dc5fca24e185629c14a

sourcebundle/builder.go
Maed223: Included 👍 https://github.com/hashicorp/go-slug/pull/59/commits/be5d90781af6a49d0b5f5dc5fca24e185629c14a

sourcebundle/builder.go
Maed223: This comes from the approach we've taken with how we denote deprecated versions. In essence, when there are no deprecations, the registry api doesn't include it in the Module Version response. So in this way`nil` denotes an un-deprecated module.

PR #67, opened 2024-10-30
sourcebundle: Fix panic when RegistryMeta and Packages aren't the same size

liamcervante:
Seems like a copy-paste error here. I think this works most of the time because packages and registrymeta are usually the same size. But, if someone loads a package directly from Github (or another source) without going through the registry this panics.


radeksimko (approved):
Agreed this looks like a copy-paste error.

I guess it is not very surprising for something like this to be overlooked in 4,7k+ LOC of changes https://github.com/hashicorp/go-slug/pull/42 🤷🏻

liamcervante:
Closed in favour of https://github.com/hashicorp/go-slug/pull/68

PR #74, opened 2024-12-02
Deprecate unused pack/unpack option `AllowSymlinkTarget`

NodyHub:
Deprecate option to inform users that the option will be removed in a future release of the library. The motivation is to simplify the implementation logic and prepare the adoption of [hashicorp/go-extract](https://github.com/hashicorp/go-extract).


dduzgun-security (approved):
LGTM 👍

PR #76, opened 2025-01-06
Adjust destination path check

NodyHub:
Extend the check that verifies that every element in the destination path is NOT a symlink. Additionally, ensure that the extraction stays in the destination directory.


dduzgun-security (approved):
LGTM 👍

mohanmanikanta2299 (approved):

PR #77, opened 2025-01-24
fix: bug in prefix and suffix checks

dduzgun-security:
Resolves a bug where the prefix check doesn't work on empty path.


dduzgun-security:
Not working as expected

PR #78, opened 2025-01-24
fix: bug in prefix and suffix check using filepath

dduzgun-security:
Alternative to https://github.com/hashicorp/go-slug/pull/77 but using`"path/filepath"` instead of strings. 
Resolves a bug where the prefix check doesn't work on empty path.

Testing with real tarballs ([ref](https://github.com/hashicorp/security-vulnerabilities/pull/362))

### DIFF
--- a/internal/unpackinfo/unpackinfo.go
+++ b/internal/unpackinfo/unpackinfo.go
@@ -35,19 +35,27 @@ func NewUnpackInfo(dst string, header *tar.Header) (UnpackInfo, error) {
 		return UnpackInfo{}, errors.New("empty destination is not allowed")
 	}
 
+	// Clean the destination path
+	dst = filepath.Clean(dst)
+
 	// Get rid of absolute paths.
 	path := header.Name
-
 	if filepath.IsAbs(path) {
 		path = path[1:]
 	}
 
 	path = filepath.Join(dst, path)
-
 	target := filepath.Clean(path)
+
+	// Check for path traversal by ensuring the target is within the destination
 	rel, err := filepath.Rel(dst, target)
 	if err != nil || strings.HasPrefix(rel, "..") || filepath.IsAbs(rel) {
 		return UnpackInfo{}, errors.New("invalid filename, traversal with \"..\" outside of current directory")
+	}
+
+	// Additional check for path traversal in the destination and target
+	if strings.Contains(dst, "..") || strings.Contains(target, "..") {
+		return UnpackInfo{}, errors.New("invalid path, traversal with \"..\" is not allowed")
 	}
 
 	// Ensure the destination is not through any symlinks. This prevents
--- a/internal/unpackinfo/unpackinfo.go
+++ b/internal/unpackinfo/unpackinfo.go
@@ -37,19 +37,14 @@ func NewUnpackInfo(dst string, header *tar.Header) (UnpackInfo, error) {
 
 	// Clean the destination path
 	dst = filepath.Clean(dst)
-
-	// Get rid of absolute paths.
-	path := header.Name
-	if filepath.IsAbs(path) {
-		path = path[1:]
-	}
+	path := filepath.Clean(header.Name)
 
 	path = filepath.Join(dst, path)
 	target := filepath.Clean(path)
 
 	// Check for path traversal by ensuring the target is within the destination
 	rel, err := filepath.Rel(dst, target)
-	if err != nil || strings.HasPrefix(rel, "..") || strings.Contains(dst, "..") || strings.Contains(target, "..") || filepath.IsAbs(rel) {
+	if err != nil || strings.HasPrefix(rel, "..") || strings.Contains(dst, "..") || strings.Contains(path, "..") || strings.Contains(target, "..") {
 		return UnpackInfo{}, errors.New("invalid filename, traversal with \"..\" outside of current directory")
 	}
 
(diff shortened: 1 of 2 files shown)


internal/unpackinfo/unpackinfo.go
```diff
@@ -30,20 +30,26 @@ type UnpackInfo struct {
...
-	if len(dst) > 0 && !strings.HasSuffix(dst, "/") {
-		dst += "/"
-	}
+	path = filepath.Join(dst, path)
 	target := filepath.Clean(path)
-	if !strings.HasPrefix(target, dst) {
+
+	// Check for path traversal by ensuring the target is within the destination
+	rel, err := filepath.Rel(dst, target)
+	if err != nil || strings.HasPrefix(rel, "..") || strings.Contains(dst, "..") || strings.Contains(target, "..") || filepath.IsAbs(rel) {
```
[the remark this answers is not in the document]
dduzgun-security: Ah yes that's true, I forgot to clean that check while doing tests https://github.com/hashicorp/go-slug/pull/78/commits/be79e050e98e2537b17b46e31bddb012c59af542 should remove duplicated checks.

internal/unpackinfo/unpackinfo.go
```diff
@@ -30,20 +30,26 @@ type UnpackInfo struct {
...
+
+	// Clean the destination path
+	dst = filepath.Clean(dst)
+
 	// Get rid of absolute paths.
 	path := header.Name
-
-	if path[0] == '/' {
-		path = strings.TrimPrefix(path, "/")
+	if filepath.IsAbs(path) {
```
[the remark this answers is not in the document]
dduzgun-security: Good point, this is indeed not needed for the path since the join will pop the additional remaining `/`s

internal/unpackinfo/unpackinfo.go
james-warren0: I would imagine there may be valid use cases for `dst` to contain `..`. Would checking `err != nil || strings.HasPrefix(rel, "..") || strings.Contains(target, "..")` cover the appropriate cases but still allow a potential legitmate `dst` that contains `..`?

dduzgun-security: I think that allowing the destination to contain `..` would let someone write into an escaped destination. For instance, `slug.Unpack(fileToUnpack, "../../../../../../../Users/user_name/.ssh")`

eastebry (approved):

internal/unpackinfo/unpackinfo.go
picatz: I'd like to re-raise this potential concern, just to be sure. I understand that `..` allows upward path traversal, which is obviously dangerous. 

But, if a user's file contained `..` in it ( `./docs..md` for example  ) that's not a path traversal, it could just be a weird file name. It''ll cause an error in this case, because we're using `strings.Contains` instead of just `strings.HasPrefix`.

Looking at how we handle this in Nomad, we *only* do `strings.HasPrefix`:
https://github.com/hashicorp/nomad/blob/b7aabb11be1c5b7616d82f35553cce8b01cc187b/helper/escapingfs/escapes.go#L107-L119

 So, to me, the `strings.Contains` check seems suspect, possibly unnecessary. 🤔

internal/unpackinfo/unpackinfo.go
```diff
@@ -30,20 +30,21 @@ type UnpackInfo struct {
...
+	dst = filepath.Clean(dst)
+	path := filepath.Clean(header.Name)
+
+	path = filepath.Join(dst, path)
 	target := filepath.Clean(path)
-	if !strings.HasPrefix(target, dst) {
+
+	// Check for path traversal by ensuring the target is within the destination
+	rel, err := filepath.Rel(dst, target)
+	if err != nil || strings.HasPrefix(rel, "..") || strings.Contains(dst, "..") || strings.Contains(path, "..") {
```
picatz, line 47: Following up on my other comment, this would align us with Nomad's logic, which seems sufficient / simpler?

https://github.com/hashicorp/nomad/blob/b7aabb11be1c5b7616d82f35553cce8b01cc187b/helper/escapingfs/escapes.go#L107-L119

```suggestion
	if err != nil || strings.HasPrefix(rel, "..") {
```

dduzgun-security: Good point, also thanks for reviewing it on Slack 🙌  Since the `dst` is set by us and is not a user provided data I'll go ahead and approve the suggestion and edit the destination related tests

internal/unpackinfo/unpackinfo.go
dduzgun-security: https://github.com/hashicorp/go-slug/pull/78/commits/9bf13d2b7e62eed3d6372a4d06953ceb30c39957 removes the strings.Contains check since we don't rely on user provided `dst` and it adds more failure point as described for the potential filename containing `file..ext`.

james-warren0 (approved):

picatz (approved):

dduzgun-security:
@hashicorp/team-ip-compliance when you get some time a review would be appreciated 🙏

eastebry (approved):

lafentres (approved):

internal/unpackinfo/unpackinfo.go
```diff
@@ -30,20 +30,21 @@ type UnpackInfo struct {
...
 	}
-	path = filepath.Join(dst, path)
 
-	// Check for paths outside our directory, they are forbidden
-	if len(dst) > 0 && !strings.HasSuffix(dst, "/") {
-		dst += "/"
-	}
+	// Clean the destination path
+	dst = filepath.Clean(dst)
+	path := filepath.Clean(header.Name)
```
mohanmanikanta2299, line 40: Hey @dduzgun-security,

The previous code before changes to the file had a high vulnerability issue - SECVULN-17966
I think the same problem might still exist. 

Can we plan to remediate it in the same PR or do we need to plan a spearate PR ?

PR #79, opened 2025-01-27
irregular mode file checks for Windows symlinks

notchairmk:
Adds checks for `os.ModeIrregular` anywhere where we were checking symlinks. This is required as of Go 1.23, additional discussion in golang/go#63703

### DIFF
--- a/slug.go
+++ b/slug.go
@@ -151,7 +151,7 @@ func (p *Packer) Pack(src string, w io.Writer) (*Meta, error) {
 	}
 
 	// Check if the root (src) is a symlink
-	if info.Mode()&os.ModeSymlink != 0 {
+	if isSymlink(info.Mode()) {
 		src, err = os.Readlink(src)
 		if err != nil {
 			return nil, err
@@ -257,7 +257,7 @@ func (p *Packer) packWalkFn(root, src, dst string, tarW *tar.Writer, meta *Meta,
 			header.Typeflag = tar.TypeReg
 			header.Size = info.Size()
 
-		case fm&os.ModeSymlink != 0:
+		case isSymlink(info.Mode()):
 			// Read the symlink file to find the destination.
 			target, err := os.Readlink(path)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


notchairmk:
hey @dduzgun-security what this basically boils down to is that Windows and now Go only treat _files_ as symlinks, not directories. the go issue mentioned in the description has more details on this.

in the context of go-slug, there is no distinction between symlinked directories and files. we could make this more specific like `m&os.ModeSymlink != 0 || (m&os.ModeIrregular != 0 && runtime.GOOS == 'windows')` or by adding comments describing what's going on. but it doesn't feel like making this check separate from the `os.ModeSymlink` checks is preferable.

notchairmk:
also just want to note that tests for Windows will fail on main already, they're not necessarily due to this change.

dduzgun-security:
@notchairmk Thanks for the context, I would suggest a condition for Windows first by using the check `runtime.GOOS == 'windows'` then following up with the symlink & irregular file check to reduce the impact of the change. This library is quite sensitive and used by many, targeting the change for the affected environment would reduce the risk of incidents.

dduzgun-security (approved):
The changes looks good. I would recommend some end-to-end & unit tests to ensure nothing breaks in consumers of go-slug (eg: tfc-agent) prior to merging the PR.

dduzgun-security:
The PR might also need a rebase prior to merging

sonamtenzin2 (dismissed):

Maed223:
I haven't been able to directly validate the fix here due to issues with getting a proper windows testing environment setup. Though there's a guaranteed fix here, which is just disabling the updated Windows symlink behavior by setting `godebug` option to `winsymlink=0`. See the same fix done [in Terraform itself](https://github.com/hashicorp/terraform/pull/36575).

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/hashicorp/go-slug
 
 go 1.24
 
+godebug winsymlink=0
```
sonamtenzin2, line 5: Comment: I dont think this will work in case go-slug is used as a dependency

https://go.dev/doc/modules/gomod-ref#godebug-notes

`GODEBUG settings only apply for builds of main packages and test binaries in the current module. They have no effect when a module is used as a dependency.`

PR #80, opened 2025-01-31
fix: archive extraction header.Linkname and old dependencies

mohanmanikanta2299:
**This PR is to fix the CVEs in below tickets:**

- https://hashicorp.atlassian.net/browse/SECVULN-9879
- https://hashicorp.atlassian.net/browse/SECVULN-9881
- https://hashicorp.atlassian.net/browse/SECVULN-17660

## Note
❗ To merge and release a new version only when ProdSec and TF team have approved.


slug.go
```diff
@@ -507,19 +510,24 @@ func (p *Packer) validSymlink(root, path, target string) (bool, error) {
...
 	}
 
 	// Target falls within root.
-	if strings.HasPrefix(absTarget, absRoot) {
+	rel, err := filepath.Rel(absRoot, absTarget)
+	if err != nil {
+		return false, fmt.Errorf("couldn't find relative path : %w", err)
+	}
+
+	if rel != ".." && !(len(rel) >= 3 && rel[:3] == "../") {
```
dduzgun-security, line 530: I think we can remove this condition since the `rel` will already be cleaned.

dduzgun-security: It would be a good idea to copy the validation function the Nomad team worked on and use that to reduce code. https://github.com/hashicorp/nomad/blob/main/helper/escapingfs/escapes.go#L110-L119 maybe under `/internal/helper`

slug.go
```diff
@@ -539,7 +548,13 @@ func (p *Packer) validSymlink(root, path, target string) (bool, error) {
...
 			prefix += "/"
 		}
-		if strings.HasPrefix(absTarget, prefix) {
+
+		rel, err := filepath.Rel(prefix, absTarget)
+		if err != nil {
+			return false, fmt.Errorf("couldn't find relative path : %w", err)
+		}
+
+		if rel != ".." && !(len(rel) >= 3 && rel[:3] == "../") {
```
dduzgun-security, line 557: Similar to the following comment: https://github.com/hashicorp/go-slug/pull/80/files#r1937639814

dduzgun-security (changes requested):
Thanks a lot for looking into that, really appreciate the work. Left some comments and it would be nice to add lots of unit tests (expected failure and valid ones) covering different scenarios included.

slug.go
mohanmanikanta2299: Made the necessary changes and created a validation function as suggested. Took Nomad's example from the link shared above.

slug.go
mohanmanikanta2299: Made the necessary changes and created a validation function as suggested. Took Nomad's example from the link shared above.

slug.go
```diff
@@ -552,6 +568,17 @@ func (p *Packer) validSymlink(root, path, target string) (bool, error) {
 	}
 }
 
+func TargetWithinRoot(root string, target string) (bool, error) {
```
dduzgun-security, line 571: Should we move this to `/internal/escapingfs/escapes.go`? I'm sure this would be useful for future validations outside of the `slug.go` file too

slug.go
```diff
@@ -539,8 +548,15 @@ func (p *Packer) validSymlink(root, path, target string) (bool, error) {
 		if !strings.HasSuffix(prefix, "/") {
```
dduzgun-security, line 548: String manipulation here after a clean may cause some issues that we faced before that was resolved by https://github.com/hashicorp/go-slug/commit/d8b1ee7171000f15726d9e0583d88132232b9229. The `filepath.Join` function should already handle missing `"/"`. Some tests would be helpful on here for this.

slug.go
mohanmanikanta2299: Do we need to create a new file in the said location ?
/internal/escapingfs/escapes.go -- cannot find any existing file or filepath mentioned.

mohanmanikanta2299: Made the changes. Created a new file and a test file.

slug.go
dduzgun-security: https://github.com/hashicorp/go-slug/pull/80/commits/63afbb8c6015d9f3ebdee18902886742ef0170f6 should cover it. Since the change introduces `filepath.Join` we don't need the additional prefix.

alanpchua (approved):
LGTM

lafentres (approved):
i tested this out with a local build of the agent using two basic configs and everything works as expected. customers use symlinks in loads of different ways like cj said so my testing was definitely not comprehensive but i can confirm the cases that failed last time should not fail with this change

mallikabandaru (approved):

PR #82, opened 2025-02-28
feat: experimental capture userinfo portion when parsing remote sourc…

mjyocca:
…e address


sourceaddrs/package_remote.go
```diff
@@ -15,6 +15,9 @@ type RemotePackage struct {
 	// all relevant fields are comparable, so it's safe to compare
 	// RemotePackage using the == operator.
 	url url.URL
+
+	// NOTE: If a source address contains "userinfo" portion, it'll be captured here
+	userInfo url.Userinfo
```
mjyocca, line 20: This would break RemotePackage comparison, unless creating a helper to discard/ignore this field.

PR #86, opened 2025-03-25
IND-2752 Linting

KaushikiAnand:
- I have added linting to go-tests.yml. .go-version file was added.
- The linting was done to ensure that the codebase is consistent and maintainable and helps in improving the code quality and reducing errors. The .go-version file was added to find the security vulnerabilities in a repo.


abhijeetviswa (changes requested):

slug.go
```diff
@@ -468,7 +468,7 @@ func (p *Packer) Unpack(r io.Reader, dst string) error {
 			// has perms that don't allow overwriting. The file permissions will be restored
 			// once the file contents are copied.
 			if os.IsPermission(err) {
-				os.Chmod(info.Path, 0600)
+				_ = os.Chmod(info.Path, 0600)
```
mohanmanikanta2299, line 471: I think the error here should not be supressed, rather it should be flagged in case of failure to change the file permission.

C.c- @dduzgun-security

dduzgun-security: I agree, similar thing for the other parts in the PR too

PR #88, opened 2025-04-14
Copy Method update to avoid DoS via decompression bomb

mohanmanikanta2299:
Updated the code by using io.CopyN method instead of io.Copy to mitigate the [vulnerability](https://hashicorp.atlassian.net/browse/SECVULN-7827).

This change will help to copy the contents of the file in reasonably sized chunks by using the io.CopyN method in a loop. Added logic to limit the number of chunks copied in this loop in order to prevent an excessive number of iterations that could lead to denial of service.


internal/copyUtil/limitCopy.go
```diff
@@ -0,0 +1,38 @@
+package copyUtil
+
+import (
+	"errors"
+	"io"
+)
+
+const (
+	chunkSize = 4 * 1024 * 1024 // 4MB per chunk
+	maxChunks = 100             // Up to 400MB
```
dduzgun-security, line 10: I think for the purpose of go-slug where it's gzips, we could reduce the sizing.
```suggestion
	chunkSize = 1* 1024 * 1024 // 1MB per chunk
	maxChunks = 50             // Up to 50MB
```

internal/copyUtil/limitCopy.go
```diff
@@ -0,0 +1,38 @@
...
+				break // Copy complete
+			}
+			return err
+		}
+
+		if n < chunkSize {
+			break // No more data left
+		}
+
+		totalChunks++
```
dduzgun-security, line 34:
```suggestion

```

internal/copyUtil/limitCopy.go
```diff
@@ -0,0 +1,38 @@
...
+// copyWithLimit copies from a tar.Reader to a file in limited-size chunks
+func CopyWithLimit(dst io.Writer, src io.Reader) error {
+	var totalChunks int
+
+	for {
+		if totalChunks >= maxChunks {
+			return errors.New("copy limit exceeded")
+		}
+
+		n, err := io.CopyN(dst, src, chunkSize)
```
dduzgun-security, line 22:
```suggestion
		n, err := io.CopyN(dst, src, chunkSize)
		totalChunks++
```

dduzgun-security (changes requested):
Seems like the counter is incremented only after a full chunk is copied which could allow going over the max of 400MB by 4MB. Also, the range for the sizing seems a bit excessive for the files go-slug deals with, we could try with 50MB as a max.

dduzgun-security:
Also, was `func CopyN(dst Writer, src Reader, n int64) (written int64, err error)` not working?

slug.go
```diff
@@ -478,7 +479,7 @@ func (p *Packer) Unpack(r io.Reader, dst string) error {
 		}
 
 		// Copy the contents of the file.
-		_, err = io.Copy(fh, untar)
+		err = copyUtil.CopyWithLimit(fh, untar)
```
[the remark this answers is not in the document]
mohanmanikanta2299: Existing test are failing with this approach.

internal/copyUtil/limitCopy.go
mohanmanikanta2299: Updated as suggested.

slug.go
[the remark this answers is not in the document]
dduzgun-security: I think something like this should work
```go
// Copy the contents of the file under 10MB.
_, err = io.CopyN(fh, untar, 10*1024*1024)
fh.Close()
if err != nil && err != io.EOF {
    return fmt.Errorf("failed to copy slug file %q: %w", info.Path, err)
}
```

eastebry:
Hey folks! Just a heads-up, I think we need to be pretty careful with this change. This code path is used for extracting slugs within Terraform, and it seems plausible that these will exceed 10MB.

Generally speaking, I'm guessing the risk here is fairly low because (as far as I know, I could be wrong), slugs are mostly unpacked in single-tenant environments (either in Terraform core, or within the sandbox environment of Terraform cloud). Given this, I think we need to carefully consider if this security improvement is worth the potential user impact/risk.

Before merging this, can we please do the following:
* Talk with all consumers of go-slug (Terraform code, registry, cloud, and potentially more) to discuss what would be a reasonable limit, or if any limit we would impose could pose problems .
* If we enforce a specific limit, ensure that it is explained in our product documentation, and called out in any release notes.

dduzgun-security:
Considering all the implications here and the customers use cases which could go from unpacking files of MB to GB in size based out of feedbacks from the TF team, I think we should close/dismiss this issue to not cause any incident. Setting a hard limit would cause potential problems in the future for specific customer use case. I'll go ahead an close this alert.

PR #98, opened 2025-06-16
fix lint errors and run linter as part of acceptance checks

abhijeetviswa:


slug.go
```diff
@@ -480,7 +480,7 @@ func (p *Packer) Unpack(r io.Reader, dst string) error {
 
 		// Copy the contents of the file.
 		_, err = io.Copy(fh, untar)
-		fh.Close()
+		_ = fh.Close()
```
ritikrajdev, line 483: just out of curiosity, why are we not using `defer` over here?

ritikrajdev (dismissed):
LGTM

slug.go
ritikrajdev: Great!!

PR #100, opened 2025-07-11
SECVULN-7809 : Fix HasPrefix usage

mohanmanikanta2299:
## Description
- Use of strings.HasPrefix in path traversal checks for parents without a trailing slash. Issue is fixed using other code alternatives instead of using HasPrefix.

## Related Issue
- https://hashicorp.atlassian.net/browse/SECVULN-7809

## How Has This Been Tested?
- Appropriate test cases have been added in the places where the HasPrefix is used and tested with both old and new code.
Old code failed the tests whereas new code passes the tests.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       with:
         # We must fetch at least the immediate parents so that if this is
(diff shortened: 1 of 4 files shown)


sonamtenzin2 (approved):

internal/escapingfs/escaping.go
```diff
@@ -14,7 +15,8 @@ func TargetWithinRoot(root string, target string) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("couldn't find relative path : %w", err)
 	}
-	if strings.HasPrefix(rel, "..") {
+	components := strings.Split(rel, string(os.PathSeparator))
+	if len(components) > 0 && components[0] == ".." {
```
NodyHub, line 19: Can we add a test that covers a path like `foo/../../../../etc/passwd`?

dduzgun-security (changes requested):

dduzgun-security (approved):

PR #101, opened 2025-08-13
Update sourceaddrs package to parse and understand Registry Components

jpogran-hashi:
### Add Component Source Address Support

This PR introduces support for a new Component Source address type in the [sourceaddrs](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) package, enabling Terraform to reference components from registries similar to modules.

- Component Source Type: New ComponentSource and ComponentSourceFinal types that support component packages from registries
- Registry Integration: Full integration with terraform-registry-address for component package parsing and validation
- Subpath Support: Components can include subpaths using the  delimiter (e.g., example.com/foo/bar//subdir)
- Version Constraints: Components support version specifications with the @ syntax (e.g., foo/bar@1.0.0)


sourceaddrs/source.go
CreatorHead, line 1: NIT: Add a package-level note that we now support components (with `@version` and `//subdir`) in addition to modules. That helps downstream users.

sourceaddrs/source_component.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return ComponentSource{
+		pkg:     pkgOnlyAddr.Package,
+		subPath: subPath,
+	}, nil
+}
+
+// ParseComponentPackage parses a string into a ComponentPackage, ensuring no sub-path is present.
+func ParseComponentPackage(given string) (regaddr.ComponentPackage, error) {
```
CreatorHead, line 47: **Suggestion**: I noticed that ParseComponentPackage only checks for sub-paths, but it doesn’t stop someone from passing in a versioned address like foo/bar@1.2.3. I was just thinking in a defensive manner. Even if the parser normally strips this out, it’d be safer to explicitly reject versions here. “ComponentPackage” should always be just the package name: no sub-path and no version. Please correct me if I missed anything.

sourceaddrs/source_component.go
CreatorHead, line 1: Logic is clean and consistent overall

PR #104, opened 2025-09-09
fix(sourceaddrs): subpath handling for short-hand url expansion

mjyocca:
## Description

The source address expansion handling for shorthand URLs (GitHub, GitLab) that include a subpath is currently broken. If a valid shorthand slug includes a subpath, the user receives an incorrect format error message. This revision ensures that empty strings are excluded when building the full source address.

Example Error message
```
invalid remote source address "gitlab.com/hashicorp/stacks.git//path/to/dir?ref=v4.8.0": invalid sub-path: must be slash-separated relative path without any .. or . segments
```

### DIFF
--- a/sourceaddrs/source_remote.go
+++ b/sourceaddrs/source_remote.go
@@ -181,7 +181,7 @@ var remoteSourceShorthands = []remoteSourceShorthand{
 		if len(parts) > 3 {
 			// The remaining parts will become the sub-path portion, since
 			// the repository as a whole is the source package.
-			urlStr += "//" + strings.Join(parts[3:], "/")
+			urlStr += "//" + joinNonEmpty(parts[3:], "/")
 		}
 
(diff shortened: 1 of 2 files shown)


liamcervante (commented):
I'm concerned this will allow something like: `gitlab.com/hashicorp/stacks.git//path/to//dir`, which should be invalid right? (the second `//` isn't supported).

Is there a way to check it so that it only allows the first entry to be empty or something?

mjyocca:
> I'm concerned this will allow something like: `gitlab.com/hashicorp/stacks.git//path/to//dir`, which should be invalid right? (the second `//` isn't supported).
> 
> Is there a way to check it so that it only allows the first entry to be empty or something?

Fair and good catch. We change this to only support leading empty spaces allowing any subsequent space to be included and fall-through to our validation.

liamcervante (approved):

PR #135, opened 2026-08-24
refactor: use filepath.IsLocal for entry path validation

paulanunda:
## Description

Replace the hand-rolled archive entry path validation in `NewUnpackInfo` with `filepath.IsLocal`.

This simplifies the validation logic by using the standard library helper while preserving expected archive extraction behavior. The existing validation error message remains unchanged.

This PR also pins `golangci-lint` to v2.12.2. Resolving `latest` began enforcing an unrelated `SA4023` finding on existing `Lchtimes` calls. Upgrading the linter and addressing that finding will be handled separately.

## Related Issue

None.

## How Has This Been Tested?

`go test -race ./...` passes with the existing test coverage unchanged. `go build`, `go vet`, and `gofmt` also pass.


internal/unpackinfo/unpackinfo.go
```diff
@@ -50,12 +50,8 @@ func NewUnpackInfo(dst string, header *tar.Header) (UnpackInfo, error) {
...
 	rel = filepath.Clean(rel)
 
-	rel_components := strings.Split(rel, string(os.PathSeparator))
-
-	for _, component := range rel_components {
-		if component == ".." {
-			return UnpackInfo{}, fmt.Errorf("invalid filename, traversal with \"..\" outside of current directory")
-		}
+	if !filepath.IsLocal(rel) {
+		return UnpackInfo{}, errors.New("invalid filename, traversal with \"..\" outside of current directory")
```
jpogran-hashi:
```suggestion
		return UnpackInfo{}, errors.New("invalid filename, traversal outside of current directory")
```

Nit: This technically can throw for more than just `..` now, and the error text doesn't really reflect that, but not blocking comment.

jpogran-hashi (dismissed):

jpogran-hashi (dismissed):

jpogran-hashi (approved):

